PR #4, opened 2019-12-01
chore: cleanup unneeded chromium files

ryzokuken:
@cclauss I guess this would be it? WDYT? Should I also get rid of the `AUTHORS` file? It seems to be highly specific to Gerrit/Chromium tooling too, but idk if it's okay to lose that info... Maybe move it to some documentation or sth?


cclauss:
I am not a fan of moving or removing the AUTHORS file...  Our forebearers deserve credit for their work on this codebase.

ryzokuken:
@cclauss agreed! That's precisely why I was so conflicted. The rest look good? Would you mind pressing approve so I can merge without feeling bad about it?

PR #8, opened 2019-12-02
GitHub Actions: Enable os: windows-latest

cclauss:


targos:
I can reproduce the test failure on Windows.

ryzokuken:
@targos can you please rebase this? If not, I'd be happy tp.

targos:
I checked with my computer and it seems that part of the code is really broken on recent versions of Visual Studio. The registry values that are checked do not exist on my machine.
Related: https://developercommunity.visualstudio.com/content/problem/2813/cant-find-registry-entries-for-visual-studio-2017.html

cclauss:
I am not a Windows user so it is difficult for me to fix this so please try to make progress on it.  Given the simple mods, please feel free to create another PR and close this one if that is easier.

ryzokuken:
@targos could you please look into this issue? Would it take a non-trivial amount of effort to fix? It seems like the upstream PR at https://github.com/nodejs/node/pull/32698 is blocked on this.

targos:
I have looked into this and it requires work that I can't do myself.

This is not related to the error we saw in your PR

ryzokuken:
@targos it's not related but I think we require CI on Windows in order to fix that issue.

vweevers:
https://github.com/nodejs/node-gyp/pull/1762#issuecomment-498408442 suggested gyp can use `vswhere` for VS detection:

> This _\[node-gyp's visual studio detection\]_ is only needed for node-gyp AFAICT, so for Node core and other projects GYP3 could simply use VSWhere, require the C++ Workload, and pick the latest version. That would make it quite a lot simpler.

targos:
Is there a common location for `vswhere.exe`? It is not on my PATH by default and I have Visual Studio Community 2019.

targos:
OK, found it: https://github.com/microsoft/vswhere

`%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe`

targos:
I suppose we need to run something like this:

```
C:\Program Files (x86)\Microsoft Visual Studio\Installer>vswhere.exe -version [16.0,17.0) -property installationPath
C:\Program Files (x86)\Microsoft Visual Studio\2019\Community
```

targos:
Something we could do is bundle the executable with gyp-next. It is one of the recommended ways (vswhere is MIT-licensed).

/cc @joaocgreis what's the oldest version of Visual Studio that we need to support in node-gyp?
Assuming that we have `vswhere.exe` distributed with gyp-next, is it able to find all supported versions?

ryzokuken:
> Something we could do is bundle the executable with gyp-next. It is one of the recommended ways (vswhere is MIT-licensed).

Sounds like a pretty decent idea to me.

PR #14, opened 2020-03-15
chore: add pip requirements file

targos:


requirements.txt
```diff
@@ -0,0 +1,2 @@
+flake8
+pytest
```
cclauss: These are not runtime requirements so I would rename this file to `requirements_test.txt` or  `requirements_dev.txt` and make the similar change in the .yml file.

targos: done

PR #16, opened 2020-03-15
Apply autopep8 formatter

targos:


cclauss:
Lots of conflicts.  I would much prefer the Python Software Foundation's "black" formatter because is it not configurable (a feature!) like Standard and GoFmt.

targos:
Rebased.

targos:
I only added the max line length option, which also exists with the black formatter. Would the output be much different with it? I personally don't really care.

cclauss:
Let’s shift to `black` and not configure max_line_length.  It will default to max 88 chars per line for the reasons discussed in the black readme.  My sense is that the PSF is trying to establish to conical Python format and we should go with that flow.

targos:
Alternative: https://github.com/nodejs/gyp-next/pull/17

ryzokuken:
Closing this in favor of #17

PR #23, opened 2020-04-03
chore: remove trailing commas

targos:
`python -m flake8 . --select E231`

Refs: https://github.com/psf/black/issues/1288


cclauss:
Should we just blacken the whole codebase?

targos:
The codebase is already blackened.

@cclauss those were here because of a bug in black.
In the linked issue, they recommend to manually fix them: https://github.com/psf/black/issues/1288#issuecomment-600009465

PR #25, opened 2020-04-03
chore: avoid special characters in multiline strings

targos:
`python -m flake8 . --select E101,W191,W605`


ryzokuken:
Can you rebase this?

PR #26, opened 2020-04-03
chore: do not use bare 'except'

targos:
`python -m flake8 . --select E722`


cclauss:
Thanks for doing this!

`except:` --> `except Exception:` is a very minor improvement so where possible it is better use specific exceptions unless logging.  https://realpython.com/the-most-diabolical-python-antipattern

PR #36, opened 2020-05-03
meta: relicense to gyp-next contributors

ryzokuken:
/cc @targos @rvagg

### DIFF
--- a/LICENSE
+++ b/LICENSE
@@ -1,4 +1,4 @@
-Copyright (c) 2019 Ujjwal Sharma. All rights reserved.
+Copyright (c) 2020 Node.js contributors. All rights reserved.
 Copyright (c) 2009 Google Inc. All rights reserved.
 
 Redistribution and use in source and binary forms, with or without


rvagg:
Yeah ok, but I guess I'm still surprised that your name is in there as its own entry. When I look at the commits in here I don't really see a case that you should get a special claim over the code here. Maintaining gyp has been in the hands of Node.js collaborators for some time now, the codebase here has ***significant*** contributions that come in via nodejs/node-gyp and nodejs/node from a broad array of folks, so getting singled out seems like quite a bold move unless you think you can make a case for having some special level of ownership?

ryzokuken:
Not at all. When I moved the project into a personal repo, me and @cclauss were the ones working on it actively, and I added both of us to the LICENSE, more so in a bid to update the LICENSE than to take ownership in any way. @cclauss subsequently asked me to remove their name from the license later, which I did. I don't see what's the problem though, since I have now relicensed it to contributors. Are you proposing that I remove myself and Google Inc. from the document? I'm not a lawyer but I thought it was required to keep the original copyrights around...

ryzokuken:
@rvagg @targos is this better?

targos:
LGTM

PR #37, opened 2020-05-03
fix windows bug

ryzokuken:
Refs: https://chromium.googlesource.com/external/gyp/+/ab4aca868d68b9de9b20a9991bbbb5e78ab48a30%5E!
Refs: https://github.com/nodejs/node/pull/32698#discussion_r405540105

/cc @targos @cclauss

### DIFF
--- a/pylib/gyp/common.py
+++ b/pylib/gyp/common.py
@@ -364,7 +364,7 @@ def WriteOnDiff(filename):
                 dir=base_temp_dir,
             )
             try:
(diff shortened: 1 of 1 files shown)


cclauss:
The upstream commit message is `Python 3 gets rid of 'str'` so I don't know that I would trust it.

targos:
I don't know why, but the link in the op redirects to another commit. This is the right one: https://chromium.googlesource.com/external/gyp/+/ab4aca868d68b9de9b20a9991bbbb5e78ab48a30

Issue #40, opened 2020-05-12
upstream cross-compilation support for Windows on Arm


targos:
https://github.com/nodejs/node/pull/32867

PR #41, opened 2020-05-23
feat: port "add support for MSVC cross-compilation" from node

targos:
Original commit message:

    tools,gyp: add support for MSVC cross-compilation

    This change means that GYP can now generate two sets of projects: one
    exclusively for a host x64 machine and one containing a mix of x64 and
    Arm targets. The names of host targets are fixed up to end with
    _host.exe, and any actions involving them are fixed up. This allows
    compilation of Node on an x64 server for a Windows on Arm target.

Closes: https://github.com/nodejs/gyp-next/issues/40


pylib/gyp/input.py
```diff
@@ -1619,10 +1619,10 @@ def ExpandWildcardDependencies(targets, data):
                 index = index + 1
 
 
-def Unify(l):
-    """Removes duplicate elements from l, keeping the first element."""
+def Unify(list):
```
cclauss, line 1622: `list` is the name of a Python data type so it should never be used as a variable name.

test_gyp.py
```diff
@@ -229,8 +229,8 @@ def run_test(self, test, fmt, i):
             and not (stdout.endswith("NO RESULT\n"))
         ):
             print()
-            for l in stdout.splitlines():
-                print("    %s" % l)
+            for line in stdout.splitlines():
```
cclauss, line 232:
```suggestion
            print("\n".join("    %s" % line for line in stdout.splitlines()))
```

targos:
Updated. Thanks for the review

pylib/gyp/generator/msvs.py
```diff
@@ -3786,10 +3812,10 @@ def _GenerateMSBuildProject(project, options, version, generator_flags):
         content += _GetMSBuildLocalProperties(project.msbuild_toolset)
     content += import_cpp_props_section
     content += import_masm_props_section
-    if spec.get("msvs_enable_marmasm"):
+    if spec.get("msvs_enable_marmasm") or True:
```
targos, line 3815: @richard-townsend-arm I copied this from the downstream commit but it doesn't look right.

targos:
I'd like to wait for an answrer on https://github.com/nodejs/gyp-next/pull/41#discussion_r431804715 before merging.

pylib/gyp/generator/msvs.py
cclauss: LOL  :-)  That mod is not useful.  `if spec.get("msvs_enable_marmasm", True):` might be useful but it depends on how we want to deal with the case where `"msvs_enable_marmasm" not in spec`

richard-townsend-arm: Yes, I think my `git commit -av` habit has struck again! It should compile fine without `or True`, and (if not) I'll fix that.

richard-townsend-arm: A quick build on my own branch indicates that everything compiles fine without this `or True`. Feel free to remove it here and I can push a new patch to Node.js to remove it from there.

targos: Thank you for verifying! No need to push a patch to Node.js, we're going to downstream more changes from here.

ryzokuken:
@targos would you like to push a commit removing the unnecessary condition? Once that's done we could merge this :smile:

pylib/gyp/generator/msvs.py
targos, line 3815:
```suggestion
    if spec.get("msvs_enable_marmasm"):
```

targos:
@ryzokuken done

PR #44, opened 2020-06-02
gyp: fix XCode CLT version detection on Catalina

cclauss:
Copies @szekelyisz nodejs/node-gyp#2141 to this repo.

### DIFF
--- a/pylib/gyp/xcode_emulation.py
+++ b/pylib/gyp/xcode_emulation.py
@@ -1541,6 +1541,13 @@ def CLTVersion():
         except GypError:
             continue
 
+    regex = re.compile('Command Line Tools for Xcode\s+(?P<version>\S+)')
+    try:
+        output = GetStdout(['/usr/sbin/softwareupdate', '--history'])
+        return re.search(regex, output).groupdict()['version']
+    except GypError:
+        return None
+
 
 def GetStdoutQuiet(cmdlist):
     """Returns the content of standard output returned by invoking |cmdlist|.
--- a/pylib/gyp/xcode_emulation.py
+++ b/pylib/gyp/xcode_emulation.py
@@ -1543,8 +1543,8 @@ def CLTVersion():
 
     regex = re.compile(r'Command Line Tools for Xcode\s+(?P<version>\S+)')
     try:
-        output = GetStdout(['/usr/sbin/softwareupdate', '--history'])
-        return re.search(regex, output).groupdict()['version']
+        output = GetStdout(["/usr/sbin/softwareupdate", "--history"])
+        return re.search(regex, output).groupdict()["version"]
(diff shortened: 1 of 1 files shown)


rvagg:
@szekelyisz: any idea what `/usr/sbin/softwareupdate --history` does with software that's since been removed? Is it possible that this will find an old CLI tools that is no longer there?

szekelyisz:
> @szekelyisz: any idea what `/usr/sbin/softwareupdate --history` does with software that's since been removed? Is it possible that this will find an old CLI tools that is no longer there?

Hmm I didn't test that. AFAIK there's no (clean) way to roll back any update in macOS, and the package is gone, so the only way I can imagine to remove the tools is to delete the installation directory (usually `/Library/Developer/CommandLineTools/`). But in that case this patch will do exactly the same as happened before (compilation will fail with `command not found` or something).

pylib/gyp/xcode_emulation.py
```diff
@@ -1541,6 +1541,13 @@ def CLTVersion():
         except GypError:
             continue
 
+    regex = re.compile(r'Command Line Tools for Xcode\s+(?P<version>\S+)')
+    try:
+        output = GetStdout(['/usr/sbin/softwareupdate', '--history'])
+        return re.search(regex, output).groupdict()['version']
```
targos, line 1547:
```suggestion
        output = GetStdout(["/usr/sbin/softwareupdate", "--history"])
        return re.search(regex, output).groupdict()["version"]
```

I think `black` would add double quotes here. Could we lint for it?

BethGriggs:
As @richardlau mentioned, I had seemingly related issue with Xcode on Catalina yesterday. 

Before: 
```
creating config.gypi
creating config.status
creating config.mk
running: 
    python tools/gyp_node.py --no-parallel -f make-mac
No receipt for 'com.apple.pkg.CLTools_Executables' found at '/'.
No receipt for 'com.apple.pkg.DeveloperToolsCLILeo' found at '/'.
No receipt for 'com.apple.pkg.DeveloperToolsCLI' found at '/'.
gyp: No Xcode or CLT version detected!
Error running GYP
```

After applying the patch from https://github.com/nodejs/node-gyp/pull/2141 I was able to build (this was v10.x). The `No receipt for ...` warnings still appeared in the output though. I ended up just `sudo rm -r /Library/Developer/CommandLineTools` and reinstalling.

lorenzopicoli:
When can we expect to have this merged and released?

ryzokuken:
I'll merge this now since @BethGriggs has confirmed that it fixes the problem atleast to some extent. At some point in the near future, either me or @targos would make a release and update it in nodejs/node and nodejs/node-gyp.

Issue #47, opened 2020-06-26
copy_cmd `cp -a` isn't supported on Solaris


cjbj:
The `cp` option `-a` in copy_cmd fails on Solaris.  Currently `npm install` of a binary add-on gives:

```
cp: illegal option -- a
Usage: cp [-f] [-i] [-p] [-@] [-/] f1 f2
       cp [-f] [-i] [-p] [-@] [-/] f1 ... fn d1
       cp -r|-R [-H|-L|-P] [-f] [-i] [-p] [-@] [-/] d1 ... dn-1 dn
```

A work around is to use `node-gyp configure`, edit build/Makefile and change the `cp` syntax to use `-pPR`.  Then run `node-gyp build`:

```
bash-4.4$ diff build/Makefile build/Makefile.bak
134c134
< cmd_copy = rm -rf "$@" && cp -pPR "$<" "$@"
---
> cmd_copy = rm -rf "$@" && cp -af "$<" "$@"
```

Also see https://github.com/nodejs/node-gyp/issues/562#issuecomment-620623044

* **Node Version**: 12.18.1
* **Platform**: solaris
* **Compiler**: gcc (GCC) 7.3.0
* **Module**: oracledb

richardlau:
I've transferred this over to gyp-next.

ryzokuken:
@cjbj interesting. I just glossed over https://docs.oracle.com/cd/E88353_01/html/E37839/cp-1.html and you're indeed correct, this is a bug. That said, I'm a little unconvinced if you can just simply replace `-a` with `-pP` since `-a` implies `--preserve=all` on Linux while `-p` would fail silently on Solaris for things like mode, xattrs and timestamps. But since Solaris doesn't seem to allow explicitly specifying `--preserve`, I suppose something like `-pP@/` should work better though, maybe just `-pP@`.

> maybe add an f too?

Right, we also need to keep the `-f` unless there's a good reason for dropping it?

cjbj:
For background, I don't personally use Solaris and the `-pPR` options were given to me; they solved the immediate problem of the user who reported it.  I wouldn't expect them to cover all cases node-gyp was trying to solve.  I can see from https://github.com/nodejs/node-gyp/issues/562#issuecomment-620623044 that some node-gyp expertise is needed to make sure the final option choice is comprehensive.

ryzokuken:
[...]

PR #48, opened 2020-06-29
solaris: fix copy archive bug

ryzokuken:
Fix a bug that occurs while copying archives on solaris, because cp(1)
in solaris doesn't support the -a flag. Instead, this commit replaces
that flag with -pPR@, which should be closest to the intended behavior.

Fixes: https://github.com/nodejs/gyp-next/issues/47

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 
 ## [Unreleased]
 
+- Fixed a bug on Solaris where copying archives failed.
+
 ## [0.3.0] - 2020-06-06
 
 ### Added
--- a/pylib/gyp/generator/make.py
+++ b/pylib/gyp/generator/make.py
@@ -2362,7 +2362,14 @@ def GenerateOutput(target_list, target_dicts, data, params):
             }
         )
     elif flavor == "solaris":
-        header_params.update({"flock": "./gyp-flock-tool flock", "flock_index": 2})
+        copy_archive_arguments = "-pPRf@"
+        header_params.update(
+            {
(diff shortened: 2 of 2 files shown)


targos:
I'd like confirmation from @cjbj that it fixes their problem and maybe a look from @nodejs/platform-smartos (not sure if the platform hits this condition).

cjbj:
@targos I'll see if the developer still has the environment available.

cjbj:
From my Solaris tester:
> No sure which make.py is used in npm install, so modified all three of the following make.py files with similar changes
bash-4.4$ find . -name "make.py"
./deps/npm/node_modules/node-gyp/gyp/pylib/gyp/generator/make.py
./tools/gyp/pylib/gyp/generator/make.py
./lib/node_modules/npm/node_modules/node-gyp/gyp/pylib/gyp/generator/make.py
bash-4.4$
>
> With these changes npm install went fine.

So this PR looks OK to merge.

ryzokuken:
Great, thanks @cjbj. I wonder if we could add tests for this somehow, but for now, merging this PR should suffice. :smile: 

@targos could you please approve whenever you have the time?

rvagg:
what's the `@` doing in here? is that a gyp or a solaris thing?

cjbj:
@rvagg I understand it's a Solaris option. From https://docs.oracle.com/cd/E88353_01/html/E37839/cp-1.html:

> –@
> Preserves extended attributes. cp attempts to copy all of the source file's extended attributes along with the file data to the destination file.

I not vouching for it personally, since I don't use Solaris.

ryzokuken:
@rvagg right. Both `@` and `/` made me very uncomfortable, but I'm not the biggest Solaris fan either. :smile:

cjbj:
Thanks everyone.

PR #49, opened 2020-07-10
build: support apple silicon (arm64 darwin) builds

MarshallOfSound:
Pairs with https://github.com/nodejs/node-gyp/pull/2165

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@ The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/).
 ## [Unreleased]
 
 - Fixed a bug on Solaris where copying archives failed.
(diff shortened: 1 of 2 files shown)


cclauss:
We will have to monitor https://blog.macstadium.com/category/macstadium-news to see when the new Apple hardware is added.

MarshallOfSound:
You can test the "build" part using any machine with the Xcode 12.0 UA preview on it, but you won't be able to ensure it actually runs / works without an apple-silicon machine somewhere.  Definitely works on my DTK though 😄

ryzokuken:
@MarshallOfSound great. I'll merge this in after another review and once that's done I suppose we could make a minor release with this. Would you mind adding this to the changelog file?

MarshallOfSound:
@ryzokuken Updated

PR #54, opened 2020-07-14
.flake8: max-complexity and ignore=C901 are mutually exclusive

cclauss:
NOTE: max-complexity should be 10, not 101!  I have never seen a complexity level so high.

Also max-line-length and ignore=E501 are mutually exclusive.  max-line-length should be 88, not 180.

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,4 +1,4 @@
 [flake8]
-max-complexity = 10
-max-line-length = 88
-extend-ignore = E203,C901,E501
+max-complexity = 101
+max-line-length = 180
+extend-ignore = E203  # whitespace before ':' to agree with psf/black


.flake8
```diff
@@ -1,4 +1,4 @@
 [flake8]
-max-complexity = 10
-max-line-length = 88
-extend-ignore = E203,C901,E501
+max-complexity = 101
```
ryzokuken, line 2: > NOTE: max-complexity should be 10, not 101! I have never seen a complexity level so high.

So now we need to refactor the code until this can be 10 again?

.flake8
```diff
@@ -1,4 +1,4 @@
...
+max-line-length = 180
```
ryzokuken, line 3: This one should be pretty easy to fix, right? Can't we use black to enforce 88?

.flake8
cclauss: A future work effort perhaps

.flake8
cclauss: Yes but black will not autofix all long lines.  For example, lines that are in triple quoted strings.

ryzokuken: Ah, I remember @targos once mentioned that to me and I volunteered to fix it but didn't actually do it, so I guess that is my fault. I'll try to look into it again.

cclauss: Running __black__ only reformats two files and only fixes 1 long line.  This leaves 62 long lines that black will not fix.

These are nice tasks of first committers or interns.

cclauss: I made issues #55 and #56 which can proceed after this PR lands.

iandrc: I've already made a PR fixing [#56](https://github.com/nodejs/gyp-next/issues/56).

PR #59, opened 2020-08-05
cleanup pyc files

gengjiawen:


gengjiawen:
> These should be elimiated by [`master`/.gitignore#L3](https://github.com/nodejs/gyp-next/blob/master/.gitignore#L3) but this looks good to me.

Yeap, looks those files added before .gitignore add related rule.

PR #61, opened 2020-08-10
gyp: add missing extensions to compile_commands_json

manuel-arguelles:
Extends compile_commands_json generator to consider more file extensions
than just c and cc. This commit adds the cpp and cxx extensiosn to the
list of input source files.


cclauss:
Was nodejs/node-gyp#2189

PR #67, opened 2020-09-30
chore: sort CHANGELOG references in ASCII order

aduh95:
Refs: https://github.com/nodejs/node/pull/35423


targos:
I disagree with this change. Node.js should not validate the format of external dependencies and the order here should follow that of the headers.

aduh95:
OK, I didn't know GYP was still considered an external dependency.

PR #74, opened 2020-10-16
fix: do not rewrite absolute paths to avoid long paths

targos:
This is an alternative to f2c761838babf.

@MylesBorins @rvagg

### DIFF
--- a/pylib/gyp/generator/msvs.py
+++ b/pylib/gyp/generator/msvs.py
@@ -3614,13 +3614,12 @@ def _AddSources2(
                     extension_to_rule_name,
                     _GetUniquePlatforms(spec),
                 )
-                if group == "compile":
-                    # Always add an <ObjectFileName> value to support duplicate
-                    # source file basenames.
+                if group == "compile" and not os.path.isabs(source):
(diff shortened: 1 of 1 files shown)


gengjiawen (approved):

targos:
Tests passed upstream with the changes applied to `npm`: https://ci.nodejs.org/job/node-test-pull-request/33658/

rvagg (approved):
Does this mean absolute paths still run the risk of breaking the 260 boundary?

Since this gets us green, it's probably better than what we have now, so if you want to cut a gyp-next release, someone can put up a vendor PR to node-gyp and I can cut a release in short order. I'm off for the night but if those things can be done before tomorrow I can move fairly quickly on a release with just this change.

targos:
> Does this mean absolute paths still run the risk of breaking the 260 boundary?

No, because skipping this part for absolute paths means that we do not rename them at all, and in this case MSVC just puts the file in the output directory instead of a subdirectory.

MylesBorins:
Are we going to do an update of gyp-next and node-gyp today or should we float this on npm?

PR #75, opened 2020-10-16
chore: add release-please action

targos:
Because we used branches with names like `release-vx.y.z` for previous releases, the action should work correctly as soon as we merge this and open a release PR for the latest fix.
I adapted the changelog to the format used by `release-please`.

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -1,0 +1,16 @@
+on:
+  push:
+    branches:
+      - master
+
+name: release-please
+jobs:
+  release-please:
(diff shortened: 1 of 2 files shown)


ryzokuken (approved):
Any reason for changing the CHANGELOG format?

bcoe:
@targos awesome to see you using release-please 😄 

Let me know if you bump into any issues.

targos:
@bcoe release-please is awesome, I'm now using it in many repositories. If I had to mention one issue: the action doesn't have a `v2` tag, which means that if a bug is fixed, I have to update all my workflows to the new version.

PR #76, opened 2020-10-16
fix: only include MARMASM when toolset is target

targos:
Needed for https://github.com/nodejs/node/pull/35415 (cross-compiling for ARM64 on x64 host).


ryzokuken (approved):
LGTM

ryzokuken:
Oops, @targos I merged this without a CHANGELOG entry. Could you please add one?

targos:
With release-please, we don't need to add the changelog entries manually anymore.
See https://github.com/nodejs/gyp-next/pull/77

PR #84, opened 2020-11-29
fix(msvs): do not fix paths in action command arguments

targos:
BREAKING CHANGE: On Windows, arguments passed to the "action" commands are no longer transformed to replace slashes with backslashes.

### DIFF
--- a/pylib/gyp/generator/msvs.py
+++ b/pylib/gyp/generator/msvs.py
@@ -423,13 +423,7 @@ def _BuildCommandLineForRuleRaw(
         # file out of the raw command string, and some commands (like python) are
         # actually batch files themselves.
         command.insert(0, "call")
-        # Fix the paths
-        # TODO(quote): This is a really ugly heuristic, and will miss path fixing
-        #              for arguments like "--arg=path" or "/opt:path".
-        # If the argument starts with a slash or dash, it's probably a command line
-        # switch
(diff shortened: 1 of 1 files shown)


targos:
Node.js Windows integration is expected to fail because `js2c.py` has to take that behavior into account. https://github.com/nodejs/node/pull/36139/commits/55330331faeee42db11bdc6369f74fc7b5c20877 is the fix for it.

targos:
This change is necessary to update to V8 8.8, as it expects the paths passed to torque to always have forward slashes.

targos:
@nodejs/gyp / @nodejs/node-gyp what do you think?

gengjiawen (approved):

gengjiawen:
Anything we can do to fix windows build on this repo. Merge this will most likely breaks CI for other PR.

targos:
it's kind of a chicken and egg problem. We have to land something first either here or in nodejs/node, and CI will be broken until the other change is landed. I think it's ok to break 1 workflow here, if we don't take too much time to update node afterwards

gengjiawen:
> it's kind of a chicken and egg problem. We have to land something first either here or in nodejs/node, and CI will be broken until the other change is landed. I think it's ok to break 1 workflow here, if we don't take too much time to update node afterwards

Sounds good to me. Let's merge it :)

PR #88, opened 2020-12-30
gyp: Remove support for Python 2

cclauss:
Python 2 died one year ago so let's only support the currently supported versions of CPython:
https://devguide.python.org/#status-of-python-branches

Roughly corresponds with nodejs/node#36691 and nodejs/node-gyp#2300


targos:
I'm OK with that, but after it lands, we won't be able downstream gyp-next into Node.js until 100% of the CI is ready for Python 3

gengjiawen:
@cclauss Looks like flake8 task is failing.

gengjiawen (approved):

PR #91, opened 2021-01-16
ci: add CodeQL security scanning

smorimoto:
https://securitylab.github.com/tools/codeql

PR #92, opened 2021-01-16
GitHub Actions: Create Bandit_security_check.yml

cclauss:
https://bandit.readthedocs.io In contrast to #91 

This PR runs bandit in _allow_failures_ mode so we can see the issues without failing the build.  Perhaps the high and medium severity issues are worth looking into.
```
Code scanned:
	Total lines of code: 25393
	Total lines skipped (#nosec): 0

Run metrics:
	Total issues (by severity):
		Undefined: 0.0
		Low: 180.0
		Medium: 12.0
		High: 10.0
	Total issues (by confidence):
		Undefined: 0.0
		Low: 0.0
		Medium: 0.0
		High: 202.0
Files skipped (0):
```

### DIFF
--- a/.github/workflows/Bandit_security_check.yml
+++ b/.github/workflows/Bandit_security_check.yml
@@ -1,10 +1,10 @@
 name: Bandit_security_check
-  on: [pull_request, push]
-  jobs:
-    Bandit_security_check:
-      runs-on: ubuntu-latest
-      steps:
-        - uses: actions/checkout@v2
-        - uses: actions/setup-python@v2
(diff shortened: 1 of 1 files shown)


gengjiawen (approved):
LGTM， I think we can have both.

cclauss:
> I think we can have both.

Just because CI is free doesn't mean wasting cycles on these machines doesn't have a cost :). Our current CI load is burning down the planet.  We should not add jobs that we do not need.  Our current codebase clearly fails bandit tests.

PR #106, opened 2021-04-28
gyp: use keys in calls to sorted

nwhetsell:
This replaces uses of the `cmp` parameter of `sorted` (which was removed in Python 3) with either the `key` parameter, or no parameter where `key` seems redundant. This also encodes Unicode strings before hashing them (which is needed in Python 3).

Fixes: https://github.com/nodejs/node-gyp/issues/2101

### DIFF
--- a/pylib/gyp/xcodeproj_file.py
+++ b/pylib/gyp/xcodeproj_file.py
@@ -138,7 +138,9 @@ a project file is output.
 """
 
 import gyp.common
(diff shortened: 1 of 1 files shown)


ryzokuken (approved):

nwhetsell:
Is there anything I can do to help get this merged?

gengjiawen:
thanks for the contribution :)

PR #113, opened 2021-06-28
fix: build failure with ninja and Python 3 on Windows 

zcbenz:
There are a few python code not updated to support python 3, the code path seems to only run when using ninja on Windows.

### DIFF
--- a/pylib/gyp/generator/ninja.py
+++ b/pylib/gyp/generator/ninja.py
@@ -2389,7 +2389,6 @@ def GenerateOutputForConfig(target_list, target_dicts, data, params, config_name
         )
         if flavor == "win":
             master_ninja.variable("ld_host", ld_host)
-            master_ninja.variable("ldxx_host", ldxx_host)
         else:
             master_ninja.variable(
                 "ld_host", CommandWithWrapper("LINK", wrappers, ld_host)
--- a/pylib/gyp/msvs_emulation.py
+++ b/pylib/gyp/msvs_emulation.py
@@ -333,7 +333,7 @@ class MsvsSettings:
         # first level is globally for the configuration (this is what we consider
         # "the" config at the gyp level, which will be something like 'Debug' or
         # 'Release'), VS2015 and later only use this level
-        if self.vs_version.short_name >= 2015:
+        if int(self.vs_version.short_name) >= 2015:
             return config
         # and a second target-specific configuration, which is an
         # override for the global one. |config| is remapped here to take into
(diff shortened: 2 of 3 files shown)


pylib/gyp/generator/ninja.py
```diff
@@ -2389,7 +2389,6 @@ def GenerateOutputForConfig(target_list, target_dicts, data, params, config_name
         )
         if flavor == "win":
             master_ninja.variable("ld_host", ld_host)
-            master_ninja.variable("ldxx_host", ldxx_host)
```
targos, line 2392: `ldxx_host` is also referenced in the other branch and it doesn't seem to be an issue (I always use ninja on unix systems)

zcbenz: Python 3 fails with this error here:

```
Traceback (most recent call last):
  File "node/tools/gyp/gyp_main.py", line 45, in <module>
    sys.exit(gyp.script_main())
  File "node/tools/gyp\pylib\gyp\__init__.py", line 662, in script_main
    return main(sys.argv[1:])
  File "node/tools/gyp\pylib\gyp\__init__.py", line 654, in main
    return gyp_main(args)
  File "node/tools/gyp\pylib\gyp\__init__.py", line 639, in gyp_main
    generator.GenerateOutput(flat_list, targets, data, params)
  File "node/tools/gyp\pylib\gyp\generator\ninja.py", line 2936, in GenerateOutput
    target_list, target_dicts, data, params, config_name
  File "node/tools/gyp\pylib\gyp\generator\ninja.py", line 2393, in GenerateOutputForConfig
    master_ninja.variable("ldxx_host", ldxx_host)
UnboundLocalError: local variable 'ldxx_host' referenced before assignment
```

And that is because `ldxx_host` has never been declared for Windows:
https://github.com/nodejs/gyp-next/blob/main/pylib/gyp/generator/ninja.py#L2235-L2250

Since `ldxx_host` is never used on Windows, I think removing this line would be a reasonable fix.

gengjiawen (approved):

pylib/gyp/generator/ninja.py
gengjiawen: @targos Since the code part is in block `if flavor == "win"`, I guess it's okay ?

@zcbenz Out of curiosity, do you successfully build Node.js using Ninja on windows ?

zcbenz: > Out of curiosity, do you successfully build Node.js using Ninja on windows ?

Yes, but it requires some patches:
https://github.com/yue/node/commits/node-16.4.0

(It is used in a modified version of Node: https://github.com/yue/yode)

PR #115, opened 2021-08-09
fix: .S is an extension for asm file on Windows

zcbenz:
OpenSSL use `.S` extension for asm files and the ninja generator does not recognize it on Windows.

### DIFF
--- a/pylib/gyp/generator/ninja.py
+++ b/pylib/gyp/generator/ninja.py
@@ -1221,7 +1221,7 @@ class NinjaWriter:
                 command = "cc_s"
             elif (
                 self.flavor == "win"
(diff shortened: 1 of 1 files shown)


zcbenz:
For POSIX systems ninja uses `cc` instead of `asm` for `.S` files and it has been working fine, so I'm not touching that part.

gengjiawen (approved):

ryzokuken (approved):
Thanks!

gengjiawen:
> OpenSSL use `.S` extension for asm files and the ninja generator does not recognize it on Windows.

V8 use it too. Looks like a common pattern for asm for windows.

PR #118, opened 2021-08-13
windows related fix

gengjiawen:
Partly revert https://github.com/nodejs/gyp-next/commit/4d0f93c249286d1f0c0f665f5fe7346119f98cf1 for windows issue. See https://github.com/nodejs/node-gyp/pull/2481.

And add related github action to prevent future regression.

@cclauss @rvagg

### DIFF
--- a/.github/workflows/node-gyp.yml
+++ b/.github/workflows/node-gyp.yml
@@ -8,6 +8,8 @@ jobs:
       fail-fast: false
       matrix:
         os: [macos-latest, ubuntu-latest, windows-latest]
+        python: [3.6, 3.9]
+
     runs-on: ${{ matrix.os }}
     steps:
       - name: Clone gyp-next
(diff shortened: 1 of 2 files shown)


rvagg:
I really have no idea and am not the right person to 👍 this. I assume this fixes CI? I don't know what the implications for the cp1251 change are. @cclauss: are we risking some edge we're not accounting for properly?

ryzokuken (approved):
RSLGTM, but I agree with @rvagg that we should verify this via tests of have someone test it manually on the platform.

gengjiawen:
> I assume this fixes CI? I don't know what the implications for the cp1251 change are.

It recently add by https://github.com/nodejs/gyp-next/commit/4d0f93c249286d1f0c0f665f5fe7346119f98cf1.
But it break our node-gyp windows test. So I also add python 3.6 to ensure gyp changes will be tested on python 3.6 too.

Issue #120, opened 2021-08-16
Action defined in binding.gyp results misconfigured file path in generated vcxproj file


lkallas:
Linking to another issue where `gyp-next` is used: https://github.com/nodejs/node-gyp/issues/2485

But the problem is somewhere in the file path generation inside this repo.

lkallas:
The culprit has been found: https://github.com/nodejs/node-gyp/issues/2485#issuecomment-900844176

PR #121, opened 2021-08-22
fix(msvs): fix paths again in action command arguments

targos:
This is a revert of https://github.com/nodejs/gyp-next/pull/84
with a change to the \_FixPath function allowing to change the
separator used.

Fixes: https://github.com/nodejs/gyp-next/issues/120
Fixes: https://github.com/nodejs/node-gyp/issues/2485

### DIFF
--- a/pylib/gyp/generator/msvs.py
+++ b/pylib/gyp/generator/msvs.py
@@ -152,7 +152,7 @@ def _NormalizedSource(source):
     return source
 
 
-def _FixPath(path):
+def _FixPath(path, separator = "\\"):
     """Convert paths to a form that will make sense in a vcproj file.
 
   Arguments:
@@ -168,9 +168,12 @@ def _FixPath(path):
         and not _IsWindowsAbsPath(path)
     ):
         path = os.path.join(fixpath_prefix, path)
-    path = path.replace("/", "\\")
+    if separator == "\\":
+        path = path.replace("/", "\\")
     path = _NormalizedSource(path)
-    if path and path[-1] == "\\":
+    if separator == "/":
+        path = path.replace("\\", "/")
+    if path and path[-1] == separator:
         path = path[:-1]
     return path
 
@@ -185,9 +188,9 @@ def _IsWindowsAbsPath(path):
     return path.startswith("c:") or path.startswith("C:")
 
 
-def _FixPaths(paths):
+def _FixPaths(paths, separator):
     """Fix each of the paths of the list."""
-    return [_FixPath(i) for i in paths]
+    return [_FixPath(i, separator) for i in paths]
 
 
 def _ConvertSourcesToFilterHierarchy(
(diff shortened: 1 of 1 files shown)


pylib/gyp/generator/msvs.py
```diff
@@ -188,7 +188,7 @@ def _IsWindowsAbsPath(path):
     return path.startswith("c:") or path.startswith("C:")
 
 
-def _FixPaths(paths, separator):
+def _FixPaths(paths, separator = "\\"):
```
cclauss:
```suggestion
def _FixPaths(paths, separator="\\"):
```

targos: already fixed in the last commit

gengjiawen (approved):

targos:
@lkallas can you confirm this fixes your issue?

lkallas (approved):

lkallas:
> @lkallas can you confirm this fixes your issue?

@targos yes, with this change paths are correct and build succeeds.

lkallas:
@targos this also fixes: https://github.com/nodejs/node-gyp/issues/2399
Tried with given change and build succeeds in windows.

jrose-signal:
I *think* this is what's causing *non-*path arguments to get "../" prepended in https://github.com/signalapp/libsignal-client/runs/3657024122 (search for "../x64", see also signalapp/libsignal-client#354). Does that sound like something this commit could have caused?

This binding.gyp worked with node-gyp 7.1.2.

targos:
That's surprising. This PR essentially restored the behavior to what was in node-gyp 7.

targos:
@jrose-signal looking at another PR run in the libsignal-client repo, it looks like the command was already invoked the wrong way before, but that it was working with backslashes and is now broken because of the forward slashes? https://github.com/signalapp/libsignal-client/pull/346/checks?check_run_id=3386910497

```
      Finished release [optimized] target(s) in 3m 52s
  Invoked with '..\node\build_node_bridge.py --out-dir D:\a\libsignal-client\libsignal-client\build\Release" --os-name ..\win32 --configuration Release --cargo-build-dir Release\obj\libsignal_client_win32_x64.node\rust --cargo-target ..\x86_64-pc-windows-msvc --node-arch ..\x64'
```

I think this can be fixed on your side. Arguments that start with `/` or `-` are not modified. Can you try to change lines like `'--node-arch', '<(target_arch)'` to `'--node-arch=<(target_arch)'` ?

jrose-signal:
That's a good workaround, and I'm kicking myself for not thinking of that, but *something* has still changed since node-gyp 7. (The failure with forward slashes was an attempted run with gyp-next 8.1.0, just to see what would happen.)

targos:
> but something has still changed since node-gyp 7.

Yes. What changed is that now command arguments are "fixed" using `/` as the path separator instead of `\`.
Before: `value` would become `..\value`
Now: `value` becomes `../value`

jrose-signal:
Oh, you're right. 🤦 I misread what the build is currently doing and now I'm curious about how it's working at all. Thank you. I'll use the `=` form you suggested.

jrose-signal:
That said, I'll note it's still inconsistent with the behavior on non-Windows systems.

targos:
> That said, I'll note it's still inconsistent with the behavior on non-Windows systems.

I agree, but I tried to remove this behavior and it broke projects that relied on it :(

PR #140, opened 2022-03-03
feat: Add proper support for IBM i

kadler:
Python 3.9 on IBM i now properly returns "os400" for sys.platform
instead of claiming to be AIX as it did previously. While the IBM i PASE
environment is compatible with AIX, it is a subset and has numerous
differences which makes it beneficial to distinguish, however this means
that it now needs explicit support here.


richardlau (approved):

cclauss (approved):
Nice!  Thanks for doing this.

PR #143, opened 2022-04-06
fix(msvs): avoid fixing path for arguments with "="

targos:
They are probably not paths, and V8 uses an argument like that, so the
fix is needed in Node.js

Refs: https://github.com/nodejs/node/pull/42115


ryzokuken (approved):

cclauss (approved):
LGTM

PR #159, opened 2022-07-19
chore: Check compatibility with Python 3.11 release candidate 1

cclauss:
https://www.python.org/download/pre-releases
> Python 3.11 is up to 10-60% faster than Python 3.10

To ensure that tests are run on `pull_request` or `pull` but ***not both***, we use:
```
on:
  push:
    branches: [ main ]
  pull_request:
    branches: [ main ]
  workflow_dispatch:
```
The `workflow_dispatch` allows repo maintainers to run the tests on-demand.

### DIFF
--- a/.github/workflows/Python_tests.yml
+++ b/.github/workflows/Python_tests.yml
@@ -2,7 +2,12 @@
 # TODO: Enable pytest --doctest-modules
 
 name: Python_tests
-on: [push, pull_request]
+on:
+  push:
+    branches: [ main ]
+  pull_request:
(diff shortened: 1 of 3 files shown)


gengjiawen:
You may need downgrade windows https://github.com/nodejs/node/blob/8657d6db0703d57912e489707a8d4a40c233dfdc/.github/workflows/build-windows.yml#L37

new MSVC  has new bugs not fixed.

ryzokuken (approved):

PR #160, opened 2022-07-31
fix: decode bytes not str (easy_xml.py)

owl-from-hogvarts:
See nodejs/node-gyp/issues/2505

proper fix for #118
Special thanks to @cclauss


cclauss:
Please rebase so we can ensure that the tests pass.

owl-from-hogvarts:
> Please rebase so we can ensure that the tests pass.

Done

pylib/gyp/easy_xml.py
```diff
@@ -123,7 +123,10 @@ def WriteXmlIfChanged(content, path, encoding="utf-8", pretty=False,
 
     default_encoding = locale.getdefaultlocale()[1]
     if default_encoding and default_encoding.upper() != encoding.upper():
-        xml_string = xml_string.encode(encoding)
+        if sys.platform == "win32":
+            if isinstance(xml_string, str):
+                xml_string = xml_string.decode("cp1251")  # str --> bytes
```
legendecas, line 128: This seems incorrect to me. `XmlToString` always returns a `str` and a `str` object has no attribute `decode`.

cclauss: Correct: `str.encode()` and `bytes.decode()` exist but the opposites are undefined.
```
>>> for s_or_b in ("s", b"b"):
...     for method_name in ("encode", "decode"):
...         print(f"{s_or_b}: {method_name} is {getattr(s_or_b, method_name, None)}")
...
s: encode is <built-in method encode of str object at 0x103cbe928>
s: decode is None
b'b': encode is None
b'b': decode is <built-in method decode of bytes object at 0x103cb4bf0>
```

PR #163, opened 2022-08-26
fix: lock windows env

gengjiawen:
MSVC still exist on windows-latest, this will fix windows build.

### DIFF
--- a/.github/workflows/nodejs-windows.yml
+++ b/.github/workflows/nodejs-windows.yml
@@ -4,7 +4,7 @@ on: [push, pull_request]
 
 jobs:
   build-windows:
-    runs-on: windows-latest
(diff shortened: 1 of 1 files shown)


cclauss:
Most of the issues that Windows users open are about [windows-2022](https://github.com/actions/runner-images#available-images) and this downgrade means that we are no longer testing that environment.

targos:
We could use a matrix and test both.

ryzokuken (approved):

cclauss (approved):

gengjiawen:
> We could use a matrix and test both.

Yeap. But currently MSVC still not fixing there bug, we may have to add `allow-failure`.

PR #164, opened 2022-08-26
feat: Add command line argument for `gyp --version`

cclauss:
Support will be difficult without this basic diagnostic capability. 

DRY says we should not support the version number both in this file and `setup.py`

```
% gyp --version
v0.13.0

% gyp -V
v0.13.0

% gyp --help
usage: usage: gyp_main.py [options ...] [build_file ...]

options:
  -h, --help            show this help message and exit
  [ ... ]
  -V, --version         Show the version and exit.
```

### DIFF
--- a/pylib/gyp/__init__.py
+++ b/pylib/gyp/__init__.py
@@ -15,6 +15,9 @@ import sys
 import traceback
 from gyp.common import GypError
 
+
+VERSION = "0.13.0"
+
 # Default debug modes for GYP
 debug = {}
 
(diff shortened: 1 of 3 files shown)


pylib/gyp/__init__.py
```diff
@@ -463,8 +463,18 @@ def gyp_main(args):
...
+        "-V",
+        "--version",
+        dest="version",
+        action="store_true",
+        help="Show the version and exit.",
+    )
 
     options, build_files_arg = parser.parse_args(args)
+    if options.version:
+        print("v0.13.0")
```
gengjiawen, line 476: Can this be done reading https://github.com/nodejs/gyp-next/blob/main/setup.py ?
Currently auto release may not update this timely.

cclauss: The version number should not be stored in an executable file like `setup.py` but in a config file like `pyproject.toml` or `setup.cfg`.  I will make that change...

cclauss:
Shrink `setup.py` by adding `setup.cfg`.  https://setuptools.pypa.io/en/latest/userguide/declarative_config.html

targos:
Make sure to merge with a commit message starting with `feat:`. Then we can check if release-please works with this new setup.
For Python they need: 
image

ryzokuken (approved):

gengjiawen (approved):

PR #171, opened 2022-08-30
chore: setup.cfg: version = attr: gyp.VERSION

cclauss:
We need `release.please` to find `VERSION` in the file `pylib/gyp/__init__.py` so according to pypa/setuptools#2596
```
package_dir=
    =pylib

# plus

version = attr: gyp.VERSION
```
should give us the 1-2 punch that `release.please` needs.


targos (approved):
let's try it!

PR #179, opened 2022-10-04
fix: node.js debugger adds stderr (but exit code is 0) -> shouldn't throw

FuPeiJiang:
https://github.com/nodejs/node-gyp/pull/2719#issuecomment-1266747060

___

Debugger attached.
Waiting for the debugger to disconnect...

### DIFF
--- a/pylib/gyp/input.py
+++ b/pylib/gyp/input.py
@@ -961,13 +961,13 @@ def ExpandVariables(input, phase, variables, build_file):
                     # Fix up command with platform specific workarounds.
                     contents = FixupPlatformCommand(contents)
                     try:
-                        p = subprocess.Popen(
+                        # stderr will be printed no matter what
+                        result = subprocess.run(
                             contents,
-                            shell=use_shell,
                             stdout=subprocess.PIPE,
-                            stderr=subprocess.PIPE,
-                            stdin=subprocess.PIPE,
+                            shell=use_shell,
                             cwd=build_file_dir,
+                            check=False
                         )
                     except Exception as e:
                         raise GypError(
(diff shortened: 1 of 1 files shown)


cclauss (approved):

pylib/gyp/input.py
```diff
@@ -961,33 +961,26 @@ def ExpandVariables(input, phase, variables, build_file):
...
-                        # Simulate check_call behavior, since check_call only exists
-                        # in python 2.5 and later.
+                    if result.returncode > 0:
                         raise GypError(
                             "Call to '%s' returned exit status %d while in %s."
-                            % (contents, p.returncode, build_file)
+                            % (contents, result.returncode, build_file)
                         )
-                    replacement = p_stdout.rstrip()
+                    replacement = result.stdout.decode("utf-8").rstrip()
```
gengjiawen, line 983: In this case, maybe concat stderr info too ?

gengjiawen (approved):

pylib/gyp/input.py
FuPeiJiang: if it's merged, do I ignore this ?
___
concat stderr to what ?

to `replacement` ?
`replacement = result.stdout.decode("utf-8").rstrip() + "\n" + result.stderr.decode("utf-8").rstrip()`
the order will be changed, and I don't know how to keep original order as it appeared on screen
___
the original code did not concat stderr

gengjiawen: original code do this in `sys.stderr.write(p_stderr)`

> replacement = result.stdout.decode("utf-8").rstrip() + "\n" + result.stderr.decode("utf-8").rstrip()

Maybe put stderr first.

>if it's merged, do I ignore this ?

you can do a new PR.

FuPeiJiang: the original code is redundant:
it captures stderr,
`stderr=subprocess.PIPE,`
if stderr is not empty string
`if p.wait() != 0 or p_stderr:`
it writes it back to stderr
`sys.stderr.write(p_stderr)`
(it was only captured for condition checking)

the new code:
just don't capture it, stderr is written to stderr anyways

assuming `replacement`(stdout) is also written to stderr,
if you don't capture stderr, you don't have to concat it to stderr
___
> Maybe put stderr first.

I don't see the point of this concatenation, what does it even change ?

PR #186, opened 2023-02-18
fix: flake8 extended-ignore

moto-timo:
Fixes: Error code '#' supplied to 'ignore' option does not match '^[A-Z]{1,3}[0-9]{0,3}$'

https://stackoverflow.com/questions/74558565/flake8-error-code-supplied-to-ignore-option-does-not-match-a-z1-30

### DIFF
--- a/.flake8
+++ b/.flake8
@@ -1,4 +1,6 @@
 [flake8]
 max-complexity = 101
 max-line-length = 88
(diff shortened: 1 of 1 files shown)


cclauss (approved):

cclauss:
@moto-timo Would you be up for creating a pull request that replaces `flake8` with https://pypi.org/project/ruff ?

PR #190, opened 2023-03-02
feat(msvs): add SpectreMitigation attribute

rzhao271:
Backport of https://chromium-review.googlesource.com/c/external/gyp/+/4265144

Description from upstream:
```
Add SpectreMitigation attribute for msvs

This CL allows gyp to recognize the SpectreMitigation msvs_attribute
and add it to the generated vcxproj file for MSBuild.
Possible values for the attribute are Spectre, SpectreLoad,
SpectreLoadCF, and false.

The /Qspectre compiler option is not enough to add
full Spectre mitigation, because even though it causes MSBuild
to add additional instructions to the generated object files,
it does not cause MSBuild to link against Spectre-mitigated
libraries provided by Visual Studio.
```

### DIFF
--- a/pylib/gyp/easy_xml_test.py
+++ b/pylib/gyp/easy_xml_test.py
@@ -76,6 +76,7 @@ class TestSequenceFunctions(unittest.TestCase):
             '\'Debug|Win32\'" Label="Configuration">'
             "<ConfigurationType>Application</ConfigurationType>"
             "<CharacterSet>Unicode</CharacterSet>"
+            "<SpectreMitigation>SpectreLoadCF</SpectreMitigation>"
             "</PropertyGroup>"
             "</Project>"
         )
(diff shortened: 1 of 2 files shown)


rzhao271:
Hi @cclauss, is this PR ready to be merged?

cclauss:
I am not a Windows user so a maintainer with that knowledge will need to approve.
@nodejs/platform-windows

rzhao271:
Could a reviewer from platform-windows take a look? This PR is blocking node-gyp, and in turn, native node modules, from adopting Spectre Mitigation in accordance with [microsoft/binskim](https://github.com/microsoft/binskim) policies.

StefanStojanovic (approved):
LGTM! To add a few things I noticed while testing:
- Since VS2015 is supported [for building Node.JS Addons](https://github.com/nodejs/build/blob/main/doc/windows-visualstudio-supported-versions.md#for-building-nodejs-addons), and SpectreMitigation is available only in VS2017 and later, if VS2015 is used adding this attribute will have no effect.
- Using VS that supports SpectreMitigation (I tested on VS2017), compilation will fail if the wrong value is set or a required component is not installed. The messages in these cases are very informative, so users should know what to do.

targos (approved):

PR #202, opened 2023-07-29
Fix PLR1714 lint errors

rzhao271:
% `ruff rule PLR1714`
# repeated-equality-comparison-target (PLR1714)

Derived from the **Pylint** linter.

## What it does
Checks for repeated equality comparisons that can be rewritten as a membership test.

## Why is this bad?
To check if a variable is equal to one of many values, it is common to write a series of equality comparisons (e.g., `foo == "bar" or foo == "baz"`).

Instead, prefer to combine the values into a collection and use the `in` operator to check for membership, which is more performant and succinct.
If the items are hashable, use a `set` for efficiency; otherwise, use a `tuple`.

## Example
```python
foo == "bar" or foo == "baz" or foo == "qux"
```

Use instead:
```python
foo in {"bar", "baz", "qux"}
```

## References
- [Python documentation: Comparisons](https://docs.python.org/3/reference/expressions.html#comparisons)
- [Python documentation: Membership test operations](https://docs.python.org/3/reference/expressions.html#membership-test-operations)
- [Python documentation: `set`](https://docs.python.org/3/library/stdtypes.html#set)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,6 +46,7 @@ select = [
   "ICN",  # flake8-import-conventions
   "INT",  # flake8-gettext
   "PL",   # Pylint
+  "PLR",  # Pylint
   "PYI",  # flake8-pyi
   "RSE",  # flake8-raise
   "RUF",  # Ruff-specific rules
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -46,7 +46,6 @@ select = [
   "ICN",  # flake8-import-conventions
   "INT",  # flake8-gettext
   "PL",   # Pylint
-  "PLR",  # Pylint
   "PYI",  # flake8-pyi
   "RSE",  # flake8-raise
   "RUF",  # Ruff-specific rules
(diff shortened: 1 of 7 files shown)


cclauss:
Please adjust https://github.com/nodejs/gyp-next/blob/main/pyproject.toml#L38 to make these changes permanent.

cclauss (approved):

cclauss:
@rzhao271 The `ruff rule PLR1714` above makes it simple to make you commit message self-documenting for those who do not have all rules mesmerized.  Especially helpful for project maintainers is the `Why is this bad?` section.

pyproject.toml
```diff
@@ -46,6 +46,7 @@ select = [
   "ICN",  # flake8-import-conventions
   "INT",  # flake8-gettext
   "PL",   # Pylint
+  "PLR",  # Pylint
```
cclauss, line 49: `PL` includes all of `PLC`, `PLE`, `PLR`, and `PLW` so this change can be reverted.

cclauss (changes requested):

rzhao271:
@cclauss can the CI be restarted? The integration tests had an issue with finding the tarballs.

cclauss (approved):

PR #205, opened 2023-08-26
fix: fix PYI024 lint rule violation

rzhao271:
% `ruff rule PYI024`
# collections-named-tuple (PYI024)

Derived from the **flake8-pyi** linter.

## What it does
Checks for uses of `collections.namedtuple` in stub files.

## Why is this bad?
`typing.NamedTuple` is the "typed version" of `collections.namedtuple`.

The class generated by subclassing `typing.NamedTuple` is equivalent to
`collections.namedtuple`, with the exception that `typing.NamedTuple`
includes an `__annotations__` attribute, which allows type checkers to
infer the types of the fields.

## Example
```python
from collections import namedtuple


person = namedtuple("Person", ["name", "age"])
```

Use instead:
```python
from typing import NamedTuple


class Person(NamedTuple):
    name: str
    age: int
```


cclauss:
This rule only has power when we are type-checking with a tool like mypy.

Neither `gyp-next` nor `node-gyp` are currently running `mypy` but that would be a nice addition.
* https://github.com/search?q=repo%3Anodejs%2Fgyp-next mypy
* https://github.com/search?q=repo%3Anodejs%2Fnode-gyp+mypy

pylib/gyp/msvs_emulation.py
```diff
@@ -935,7 +935,9 @@ def BuildCygwinBashCommandLine(self, args, path_to_base):
         )
         return cmd
 
-    RuleShellFlags = collections.namedtuple("RuleShellFlags", ["cygwin", "quote"])
+    class RuleShellFlags(typing.NamedTuple):
+        cygwin: int
```
cclauss, line 939: I believe that `mypy` would agree with this change but without running `mypy` in our GitHub Actions, I would be reluctant to merge this PR.
```suggestion
        cygwin: bool
```

cclauss (changes requested):
```
cygwin = int(rule.get("msvs_cygwin_shell",
                              self.spec.get("msvs_cygwin_shell", 1))) != 0
```
Would seem to create a `bool`.

rzhao271:
Closing this PR
I tried running mypy locally and there's a lot of objects that need manual annotations

PR #213, opened 2023-10-18
nodejs-windows: python -m pip install packaging

cclauss:


targos:
Why is it needed suddenly?

rzhao271:
@cclauss PR LGTM, but I don't think it'll vendor the packaging package into the node-gyp npm package. the `pip install --editable .` line might also be installing the packaging package already, but pip can detect duplicate installs anyway, and listing out the packaging package seems clearer to me.

@targos node-gyp currently relies on distutils for detecting the Python version. Python 3.10 deprecates that package, and Python 3.12 removes distutils completely. The [recommendation](https://peps.python.org/pep-0632/#migration-advice) is to use the packaging package instead of `distutils.version`. Because node-gyp hasn't moved away from distutils yet, it is currently unusable on devices running Python 3.12.

cclauss:
Agree.  I will get the vendoring done soon.

ryzokuken (approved):

cclauss:
I put this in DRAFT because I sense that we will close it with unmerged commits in favor of #214.

PR #222, opened 2024-01-19
fix: support cross compiling for wasm with make generator

toyobayashi:
- Context: https://github.com/nodejs/node-gyp/issues/2823
- Test Repo: https://github.com/toyobayashi/emnapi-node-gyp-test

I'm not sure if it's reasonable to do such changes, but it works!

### DIFF
--- a/pylib/gyp/common.py
+++ b/pylib/gyp/common.py
@@ -428,6 +428,7 @@ def GetCrossCompilerPredefines():
     CXX = os.environ.get("CXX_target") or os.environ.get("CXX")
     CXXFLAGS = os.environ.get("CXXFLAGS")
     cmd = []
+    defines = {}
 
     if CC:
         cmd += CC.split(" ")
@@ -438,7 +439,7 @@ def GetCrossCompilerPredefines():
         if CXXFLAGS:
             cmd += CXXFLAGS.split(" ")
     else:
-        return None
+        return defines
 
     if sys.platform == "win32":
         fd, input = tempfile.mkstemp(suffix=".c")
@@ -458,7 +459,6 @@ def GetCrossCompilerPredefines():
         )
         lines = out.communicate()[0].decode("utf-8").replace(
             "\r\n", "\n").split("\n")
-        defines = {}
         for line in lines:
             if not line:
                 continue
--- a/pylib/gyp/common.py
+++ b/pylib/gyp/common.py
@@ -440,19 +440,35 @@ def GetCrossCompilerPredefines():
     else:
         return None
 
-    out = subprocess.Popen(
-        [*cmd, "-dM", "-E", "-x", "c", "/dev/null"],
-        stdout=subprocess.PIPE, stderr=subprocess.STDOUT
-    )
-    lines = out.communicate()[0].decode("utf-8").split("\n")
-    defines = {}
-    for line in lines:
-        if not line:
-            continue
-        define_directive, key, *value = line.split(" ")
-        assert define_directive == "#define"
-        defines[key] = " ".join(value)
-    return defines
+    if sys.platform == "win32":
+        fd, input = tempfile.mkstemp(suffix=".c")
+        try:
+            os.close(fd)
+        except Exception:
+            os.unlink(input)
+            raise
+    else:
+        input = "/dev/null"
+
+    try:
+        out = subprocess.Popen(
+            [*cmd, "-dM", "-E", "-x", "c", input],
+            shell=sys.platform == "win32",
+            stdout=subprocess.PIPE, stderr=subprocess.STDOUT
+        )
+        lines = out.communicate()[0].decode("utf-8").replace(
+            "\r\n", "\n").split("\n")
+        defines = {}
+        for line in lines:
+            if not line:
+                continue
+            define_directive, key, *value = line.split(" ")
+            assert define_directive == "#define"
+            defines[key] = " ".join(value)
+        return defines
+    finally:
+        if sys.platform == "win32":
+            os.unlink(input)
 
 def GetFlavorByPlatform():
     """Returns |params.flavor| if it's set, the system's default flavor else."""
--- a/pylib/gyp/common.py
+++ b/pylib/gyp/common.py
@@ -422,6 +422,37 @@ def EnsureDirExists(path):
     except OSError:
         pass
 
+def GetCrossCompilerPredefines():
+    CC = os.environ.get("CC_target") or os.environ.get("CC")
+    CFLAGS = os.environ.get("CFLAGS")
+    CXX = os.environ.get("CXX_target") or os.environ.get("CXX")
+    CXXFLAGS = os.environ.get("CXXFLAGS")
+    cmd = []
+
+    if CC:
+        cmd += CC.split(" ")
+        if CFLAGS:
+            cmd += CFLAGS.split(" ")
+    elif CXX:
+        cmd += CXX.split(" ")
+        if CXXFLAGS:
+            cmd += CXXFLAGS.split(" ")
+    else:
+        return None
+
+    out = subprocess.Popen(
+        [*cmd, "-dM", "-E", "-x", "c", "/dev/null"],
+        stdout=subprocess.PIPE, stderr=subprocess.STDOUT
+    )
+    lines = out.communicate()[0].decode("utf-8").split("\n")
+    defines = {}
+    for line in lines:
+        if not line:
+            continue
+        define_directive, key, *value = line.split(" ")
+        assert define_directive == "#define"
+        defines[key] = " ".join(value)
+    return defines
 
 def GetFlavor(params):
     """Returns |params.flavor| if it's set, the system's default flavor else."""
@@ -433,6 +464,19 @@ def GetFlavor(params):
 
     if "flavor" in params:
         return params["flavor"]
+
+    try:
+        defines = GetCrossCompilerPredefines()
+        if "__EMSCRIPTEN__" in defines:
+            return "emscripten"
+        if "__wasm__" in defines:
+            if "__wasi__" in defines:
+                return "wasi"
+            else:
+                return "wasm"
+    except Exception:
+        pass
+
     if sys.platform in flavors:
         return flavors[sys.platform]
     if sys.platform.startswith("sunos"):
--- a/pylib/gyp/common.py
+++ b/pylib/gyp/common.py
@@ -443,28 +443,32 @@ def GetCrossCompilerPredefines():  # -> dict
 
     if sys.platform == "win32":
         fd, input = tempfile.mkstemp(suffix=".c")
-        os.close(fd)
+        try:
+            os.close(fd)
+            out = subprocess.Popen(
+                [*cmd, "-dM", "-E", "-x", "c", input],
+                shell=True,
+                stdout=subprocess.PIPE, stderr=subprocess.STDOUT
+            )
+        finally:
+            os.unlink(input)
     else:
         input = "/dev/null"
-
-    try:
         out = subprocess.Popen(
             [*cmd, "-dM", "-E", "-x", "c", input],
-            shell=sys.platform == "win32",
+            shell=False,
             stdout=subprocess.PIPE, stderr=subprocess.STDOUT
         )
-        lines = out.communicate()[0].decode("utf-8").replace(
-            "\r\n", "\n").split("\n")
-        for line in lines:
-            if not line:
-                continue
-            define_directive, key, *value = line.split(" ")
-            assert define_directive == "#define"
-            defines[key] = " ".join(value)
-        return defines
-    finally:
-        if sys.platform == "win32":
-            os.unlink(input)
+
+    lines = out.communicate()[0].decode("utf-8").replace(
+        "\r\n", "\n").split("\n")
+    for line in lines:
+        if not line:
+            continue
+        define_directive, key, *value = line.split(" ")
+        assert define_directive == "#define"
+        defines[key] = " ".join(value)
+    return defines
 
 def GetFlavorByPlatform():
     """Returns |params.flavor| if it's set, the system's default flavor else."""
--- a/pylib/gyp/common.py
+++ b/pylib/gyp/common.py
@@ -450,6 +450,7 @@ def GetCrossCompilerPredefines():  # -> dict
                 shell=True,
                 stdout=subprocess.PIPE, stderr=subprocess.STDOUT
             )
+            stdout = out.communicate()[0]
         finally:
             os.unlink(input)
     else:
@@ -459,9 +460,9 @@ def GetCrossCompilerPredefines():  # -> dict
             shell=False,
             stdout=subprocess.PIPE, stderr=subprocess.STDOUT
         )
+        stdout = out.communicate()[0]
 
-    lines = out.communicate()[0].decode("utf-8").replace(
-        "\r\n", "\n").split("\n")
+    lines = stdout.decode("utf-8").replace("\r\n", "\n").split("\n")
     for line in lines:
         if not line:
             continue
--- a/pylib/gyp/common.py
+++ b/pylib/gyp/common.py
@@ -422,7 +422,7 @@ def EnsureDirExists(path):
     except OSError:
         pass
 
-def GetCrossCompilerPredefines():
+def GetCrossCompilerPredefines():  # -> dict
     CC = os.environ.get("CC_target") or os.environ.get("CC")
     CFLAGS = os.environ.get("CFLAGS")
     CXX = os.environ.get("CXX_target") or os.environ.get("CXX")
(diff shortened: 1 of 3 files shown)


pylib/gyp/generator/make.py
```diff
@@ -859,7 +859,11 @@ def Write(
             self.output = self.ComputeMacBundleOutput(spec)
             self.output_binary = self.ComputeMacBundleBinaryOutput(spec)
         else:
-            self.output = self.output_binary = self.ComputeOutput(spec)
+            if self.flavor == "win":
+                # prevent from generating copy targets on Windows
+                self.output = self.output_binary = self.ComputeOutput(spec).replace('\\', '/')
```
cclauss, line 864:
```suggestion
                self.output = self.output_binary = self.ComputeOutput(spec).replace(
                    '\\', '/'
                )
```

pylib/gyp/generator/make.py
```diff
@@ -2447,7 +2449,9 @@ def CalculateMakefilePath(build_file, base_name):
     makedep_arguments = "-MMD"
 
     # some linkers don't support --start-group/--end-group (e.g. wasm-ld)
-    link_commands = LINK_COMMANDS_LINUX.replace(' -Wl,--start-group', '').replace(' -Wl,--end-group', '') if gyp.common.CrossCompileRequested() else LINK_COMMANDS_LINUX
+    link_commands = LINK_COMMANDS_LINUX.replace(' -Wl,--start-group', '').replace(
+        ' -Wl,--end-group', ''
+    ) if gyp.common.CrossCompileRequested() else LINK_COMMANDS_LINUX
```
cclauss, line 2454: Readability...
```suggestion
    link_commands = LINK_COMMANDS_LINUX
    if gyp.common.CrossCompileRequested():
        link_commands = link_commands.replace(' -Wl,--start-group', '').replace(
            ' -Wl,--end-group', ''
        )
```

toyobayashi:
In fact the generated file path in makefile is still incorrect on Windows,  users need to use a additional script to make it work

```js
const path = require('path')
const fs = require('fs')

const buildDir = path.join(__dirname, '../build')

fs.readdirSync(buildDir)
  .filter(p => p.endsWith('.target.mk'))
  .map((p) => path.join(buildDir, p))
  .forEach((p) => {
    const content = fs.readFileSync(p, 'utf8').replace(/\\|\\\\/g, '/').replace(/\/(\r?\n)/g, '\\$1')
    fs.writeFileSync(p, content, 'utf8')
  })
```

I'm not familiar the gyp code base, it will be nice if you could help make node-gyp more friendly for building wasm🙏

toyobayashi:
> In fact the generated file path in makefile is still incorrect on Windows, users need to use a additional script to make it work

with changes in 882c1f9, the additional script is no longer required in my minimum test project, not sure if `self.WriteLn` or somewhere also need to apply the `ReplaceSep`

pylib/gyp/generator/make.py
```diff
@@ -724,6 +725,10 @@ def QuoteIfNecessary(string):
         string = '"' + string.replace('"', '\\"') + '"'
     return string
 
+def ReplaceSep(string):
```
cclauss, line 728: This is Python.  Functions are snake case and classes are CamelCase.
```suggestion
def replace_sep(string):
```

toyobayashi: Done. I just followed the naming of `QuoteIfNecessary` 😅

https://github.com/nodejs/gyp-next/blob/82d5d76bdcae91760aab25a582ad726974c99936/pylib/gyp/generator/make.py#L721

toyobayashi:
@cclauss  want to ask a question that may be not related to this PR, how can I get the absolute path of `.gypi` in itself?

The emcc `--js-library` flag can receive an absolute or relative path (relative to cwd) of a JavaScript library file to link against. If I set a relative path in `common.gypi`, 

```py
# ${projectRoot}/node_modules/emnapi/common.gypi
{
  'target_defaults': {
    'target_conditions': [
      ['_type=="executable"', {
        'conditions': [
          ['OS == "emscripten"', {
            'libraries': [
               # expect "${projectRoot}/node_modules/emnapi/dist/library_napi.js"
              '--js-library=./dist/library_napi.js',
            ],
          }],
        ],
      }],
  },
}
```


it doesn't work when running `node-gyp build` because the relative path is not handled by gyp, the cwd is `${projectRoot}` so emscripten will find the path in `${projectRoot}/dist/library_napi.js`, that is incorrect.

I tried defining a variable that invoke Node.js's `require` to locate the correct absolute path

```py
'variables': {
  'emnapi_js_library%': '<!(node -p "(()=>{try{return require(\'emnapi\').js_library}catch(e){return \'\'}})()")'
},
```

But this is not reliable, The package found by `require('emnapi')` is not necessarily the package where common.gypi is currently located, also may enter the error path if emnapi is not used via npm.

So I wonder if I can set the absolute path depend on `common.gypi` itself. So that I can write something like

```py
{
  'libraries': [
    # like CMAKE_CURRENT_SOURCE_DIR
    '--js-library=<(SOURCE_DIR_OF_COMMONGYPI)/dist/library_napi.js',
  ],
}
```

in emnapi's `common.gypi` file.

toyobayashi:
> how can I get the absolute path of .gypi in itself?

I have found solution for this question.

I opened another PR https://github.com/nodejs/node-gyp/pull/2974 in node-gyp repo to support `rebuild` and `build` commands on Windows. That and this one both are ready for review.

toyobayashi:
Any other suggestions or change request?

toyobayashi:
@cclauss Could you take a look? It's large improvement of development experience for building Node.js addon to wasm.

cclauss:
@node-gyp Are there any wasm-knowledgeable folks who would be willing to review these proposed changes?

toyobayashi:
Maybe @legendecas can review? 

Tests: https://github.com/toyobayashi/emnapi-node-gyp-test/actions

legendecas (commented):
I have no access to a Windows build machine at the moment and I will try to build the project on Windows soon.

It seems that the `replace_sep` is to replace path seperators on Windows for WSL. Does this mean that the patch no longer supports Windows builds that are not on WSL?

cclauss:
Python's https://docs.python.org/3/library/os.html#os.sep should tell you which way the slashes should lean on Windows vs. WSL.

toyobayashi:
> It seems that the replace_sep is to replace path seperators on Windows for WSL. Does this mean that the patch no longer supports Windows builds that are not on WSL?

On WSL the `sys.platform` is not `win32`, but Cygwin / MSYS is.

I don't think the make generator (`pylib/gyp/generator/make.py`) originally supports non-POSIX-like environment on Windows since it generates sh scripts and originally generates `\`, but `make` require `/` instead of `\`.

When building WebAssembly on Windows, the MSVC generator doesn't support Emscripten / wasi-sdk toolchain so makefile (`-f make-linux`) is the only choice.

This PR not only fix the Windows path problem, but also fix building `static_library` type of target in cross-compiling. If there is still any concern about breaking something, I'm ok to add  `gyp.common.CrossCompileRequested()` into `replace_sep`.

toyobayashi:
To be honest, Windows is nightmare, all these changes are the result of countless rebuild trials and errors. If you like, you can temporary revert any of these changes and then build the test project to see what will happen.

pylib/gyp/generator/make.py
```diff
@@ -2435,33 +2450,52 @@ def CalculateMakefilePath(build_file, base_name):
...
+        srcdir = replace_sep(gyp.common.RelativePath(srcdir, options.generator_output))
         srcdir_prefix = "$(srcdir)/"
 
     flock_command = "flock"
     copy_archive_arguments = "-af"
     makedep_arguments = "-MMD"
+
+    # some linkers don't support --start-group/--end-group (e.g. wasm-ld)
+    link_commands = LINK_COMMANDS_LINUX
+    if gyp.common.CrossCompileRequested():
```
legendecas, line 2462: I believe wasm-ld is the only one that doesn't support archive groups rather than if it is cross compiling. Emscripten linker supports the options as well.

Could we replace this condition with wasm-ld detection instead? `gyp.common.GetFlavor` may also be adapted to return `wasm32`, `wasm32-wasi`, or `emscripten` so that user doesn't have to define `OS` in their build scripts like https://github.com/toyobayashi/emnapi-node-gyp-test/blob/main/package.json#L10

legendecas:
Overall LGTM.

pylib/gyp/generator/make.py
toyobayashi: Done. But auto detect flavor seems doesn't work, it throws strange errors. https://github.com/toyobayashi/emnapi-node-gyp-test/actions/runs/8507038774.

> gyp.common.GetFlavor may also be adapted to return wasm32, wasm32-wasi, or emscripten so that user doesn't have to define OS

emnapi currently adds `--target` cflags according to `OS` (gyp disallow undefined variables). But auto detect can only distinguish emscripten and wasi, only when user specifies --target=wasm32 can wasi and wasm32 be further distinguished.

toyobayashi: seems not hit `OS == "emscripten"` condition on Linux and macOS. https://github.com/toyobayashi/emnapi-node-gyp-test/blob/18364bdd30cb232ba9c77344b7e4cae2da7f7c3d/binding.gyp#L14

Windows build failed: binding.target.mk:93: *** target pattern contains no '%'.  Stop.

toyobayashi: explicit `-f make-<flavor>` works. need figure out why emscripten flavor failed to detected and what happened on Windows.

pylib/gyp/common.py
```diff
@@ -422,17 +422,62 @@ def EnsureDirExists(path):
     except OSError:
         pass
 
+def GetCrossCompilerPredefines():
```
cclauss, line 425: Let's always return a list.  Functions that return `list | None` put more work on the caller.
```suggestion
def GetCrossCompilerPredefines():
```

pylib/gyp/common.py
```diff
@@ -422,17 +422,62 @@ def EnsureDirExists(path):
...
+    if CC:
+        cmd += CC.split(" ")
+        if CFLAGS:
+            cmd += CFLAGS.split(" ")
+    elif CXX:
+        cmd += CXX.split(" ")
+        if CXXFLAGS:
+            cmd += CXXFLAGS.split(" ")
+    else:
+        return None
```
cclauss, line 441:
```suggestion
    return cmd
```

pylib/gyp/common.py
```diff
@@ -422,17 +422,62 @@ def EnsureDirExists(path):
...
 
-def GetFlavor(params):
+    if sys.platform == "win32":
+        fd, input = tempfile.mkstemp(suffix=".c")
+        try:
+            os.close(fd)
+        except Exception:
```
cclauss, line 448: PEP8 warns about broad exceptions like this but we can let it slide...
* https://realpython.com/the-most-diabolical-python-antipattern

pylib/gyp/common.py
```diff
@@ -452,6 +497,24 @@ def GetFlavor(params):
...
+            return "emscripten"
+        if "__wasm__" in defines:
+            if "__wasi__" in defines:
+                return "wasi"
+            else:
+                return "wasm"
+    except Exception:
+        pass
+
+    return GetFlavorByPlatform()
```
cclauss, line 516:
```suggestion
    defines = GetCrossCompilerPredefines()
    if "__EMSCRIPTEN__" in defines:
        return "emscripten"
    if "__wasm__" in defines:
        return "wasi" if "__wasi__" in defines else "wasm"

    return GetFlavorByPlatform()
```

pylib/gyp/common.py
```diff
@@ -422,17 +422,62 @@ def EnsureDirExists(path):
...
+    if CC:
+        cmd += CC.split(" ")
+        if CFLAGS:
+            cmd += CFLAGS.split(" ")
+    elif CXX:
+        cmd += CXX.split(" ")
+        if CXXFLAGS:
+            cmd += CXXFLAGS.split(" ")
+    else:
+        return defines
```
cclauss, line 442:
```suggestion
    cmd
```

toyobayashi: GetCrossCompilerPredefines should return `defines (dict)`, not the `cmd`

pylib/gyp/common.py
toyobayashi: Done. Sorry I'm new to python 😅

pylib/gyp/common.py
toyobayashi: > defines will ALWAYS be empty.

No, `defines` is filled in

```py
    try:
        out = subprocess.Popen(
            [*cmd, "-dM", "-E", "-x", "c", input],
            shell=sys.platform == "win32",
            stdout=subprocess.PIPE, stderr=subprocess.STDOUT
        )
        lines = out.communicate()[0].decode("utf-8").replace(
            "\r\n", "\n").split("\n")
        for line in lines:
            if not line:
                continue
            define_directive, key, *value = line.split(" ")
            assert define_directive == "#define"
            defines[key] = " ".join(value)  # <---------
        return defines  # <---------
    finally:
        if sys.platform == "win32":
            os.unlink(input)
```

pylib/gyp/common.py
```diff
@@ -422,17 +422,58 @@ def EnsureDirExists(path):
...
+        )
+        lines = out.communicate()[0].decode("utf-8").replace(
+            "\r\n", "\n").split("\n")
+        for line in lines:
+            if not line:
+                continue
+            define_directive, key, *value = line.split(" ")
+            assert define_directive == "#define"
+            defines[key] = " ".join(value)
+        return defines
```
cclauss, line 464: Putting this many lines in a try block is an antipattern.

`defines` was never defined as an empty dict so line 463 will ALWAYS raise a NameError.  There is no `except` block so the code will silently jump the the `finally` block having done no useful work.  The function will then return `None` instead of a dict.

pylib/gyp/common.py
toyobayashi: > cmd may be filled up but it will ALWAYS be thrown away.

and cmd is used here

```py
try:
        out = subprocess.Popen(
            [*cmd, "-dM", "-E", "-x", "c", input], # <---------
            shell=sys.platform == "win32",
            stdout=subprocess.PIPE, stderr=subprocess.STDOUT
        )
        lines = out.communicate()[0].decode("utf-8").replace(
            "\r\n", "\n").split("\n")
        for line in lines:
            if not line:
                continue
            define_directive, key, *value = line.split(" ")
            assert define_directive == "#define"
            defines[key] = " ".join(value)
        return defines
    finally:
        if sys.platform == "win32":
            os.unlink(input)
```

pylib/gyp/common.py
toyobayashi: These code are still inside GetCrossCompilerPredefines, `defines` is defined. The diff may look misleading, I recommend viewing this code in its entirety. I can try adding test code

pylib/gyp/common.py
```diff
@@ -422,17 +422,63 @@ def EnsureDirExists(path):
...
+    if CC:
+        cmd += CC.split(" ")
+        if CFLAGS:
+            cmd += CFLAGS.split(" ")
+    elif CXX:
+        cmd += CXX.split(" ")
+        if CXXFLAGS:
+            cmd += CXXFLAGS.split(" ")
+    else:
+        return defines
```
cclauss, line 442:
```suggestion
    cmd = []
    if CC := os.environ.get("CC_target") or os.environ.get("CC"):
        cmd += CC.split(" ")
        if CFLAGS := os.environ.get("CFLAGS"):
            cmd += CFLAGS.split(" ")
    elif CXX = os.environ.get("CXX_target") or os.environ.get("CXX"):
        cmd += CXX.split(" ")
        if CXXFLAGS := os.environ.get("CXXFLAGS"):
            cmd += CXXFLAGS.split(" ")
    else:
        return {}

```

cclauss (changes requested):

pylib/gyp/common.py
```diff
@@ -422,17 +422,63 @@ def EnsureDirExists(path):
...
+    else:
+        input = "/dev/null"
+        out = subprocess.Popen(
+            [*cmd, "-dM", "-E", "-x", "c", input],
+            shell=False,
+            stdout=subprocess.PIPE, stderr=subprocess.STDOUT
+        )
+        stdout = out.communicate()[0]
+
+    lines = stdout.decode("utf-8").replace("\r\n", "\n").split("\n")
```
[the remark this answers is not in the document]
cclauss: This too.

pylib/gyp/common.py
toyobayashi: Done. Learned `:=`😄

pylib/gyp/common.py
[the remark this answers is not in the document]
toyobayashi: already changed in 3a962e9

https://github.com/nodejs/gyp-next/blob/3a962e9702594fee81c1ec36588217c17fa24d72/pylib/gyp/common.py#L459-L460

cclauss (approved):
Nice work!  Tests and all.

legendecas (approved):
Thanks!

PR #223, opened 2024-03-01
fix: avoid quoting cflag name and parameter with space separator

legendecas:
Fixes: https://github.com/nodejs/node-gyp/issues/2858

Avoid quoting `-arch arm64` in generated `compile_commands.json` as `c++ "-arch arm64"`.

### DIFF
--- a/pylib/gyp/xcode_emulation.py
+++ b/pylib/gyp/xcode_emulation.py
@@ -579,7 +579,8 @@ class XcodeSettings:
 
         sdk_root = self._SdkPath()
         if "SDKROOT" in self._Settings() and sdk_root:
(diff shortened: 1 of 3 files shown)


legendecas:
@nodejs/gyp I'd appreciate it if this could have a review, thank you!

cclauss (changes requested):
Please consider adding tests that fail without these changes but pass with these changes so that we can avoid future regressions.

cclauss (approved):
Nice!

Perhaps create a DRAFT PR with only the test changes to prove that the tests fail on the unmodified code

legendecas:
Sure, https://github.com/nodejs/gyp-next/pull/224

PR #225, opened 2024-03-07
Update Node.js in workflows and fix release-please config

targos:
- **chore: use Node.js 20 for node-gyp integration tests**
- **chore: fix release-please config**

### DIFF
--- a/.github/workflows/release-please.yml
+++ b/.github/workflows/release-please.yml
@@ -11,6 +11,4 @@ jobs:
       - uses: google-github-actions/release-please-action@v4
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          release-type: python
-          package-name: gyp-next
-          bump-minor-pre-major: true
+          config-file: release-please-config.json
--- a/.github/workflows/node-gyp.yml
+++ b/.github/workflows/node-gyp.yml
@@ -26,7 +26,7 @@ jobs:
           path: node-gyp
       - uses: actions/setup-node@v4
         with:
-          node-version: 18.x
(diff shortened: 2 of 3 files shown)


targos:
Rebased to use the pinned Ruff version.

.github/workflows/release-please.yml
```diff
@@ -11,6 +11,4 @@ jobs:
       - uses: google-github-actions/release-please-action@v4
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
-          release-type: python
-          package-name: gyp-next
-          bump-minor-pre-major: true
+          config-file: release-please-config.json
```
cclauss, line 14: Why is it useful to put the release context in two separate files?

targos: It's not useful, we have no choice as it's mandatory since the version 4 of the action.

cclauss (approved):

cclauss:
Oh… https://github.com/nodejs/gyp-next/actions/workflows/release-please.yml

targos:
Mmh, it wasn't obvious from the docs that both file must be there. I'll have a closer look.

lukekarrys:
We just did the same upgrade in `node-gyp`. I believe the only thing you will need is the top level `.release-please-manifest.json` file with the following contents:

```json
{
    ".": "0.16.2"
}
```

.github/workflows/release-please.yml
lukekarrys: You could also remove this option since that is also the default.

PR #226, opened 2024-03-07
chore: update Ruff to 0.3.1

targos:
- Fix new UP032 errors
    - Fix deprecated Ruff config structure

### DIFF
--- a/pylib/gyp/generator/android.py
+++ b/pylib/gyp/generator/android.py
@@ -739,9 +739,9 @@ class AndroidMkWriter:
                     % (self.android_class, self.android_module)
                 )
             else:
-                path = "$(call intermediates-dir-for,{},{},,,$(GYP_VAR_PREFIX))".format(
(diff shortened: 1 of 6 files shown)


legendecas (commented):
Can we pin ruff version as well?

legendecas:
FWIW, I pinned ruff to v0.3.0 in https://github.com/nodejs/gyp-next/pull/223

targos:
Ok, I changed this to a Ruff update.

legendecas (approved):

cclauss (approved):

cclauss:
My preference is that we do not pin `ruff` because it is still in a state of flux and fixing Python code incrementally in a JavaScript codebase ensures that the squeaky wheel gets the grease.  If we silence `ruff` then we will build a backlog of diverse changes that will be difficult to implement and get reviewed.

PR #229, opened 2024-03-08
Release please config updates

lukekarrys:
I think this fixes the issues with the `release-please` workflow failing after #225.

It matches mostly what we have in https://github.com/nodejs/node-gyp


.github/workflows/release-please.yml
```diff
@@ -7,8 +7,8 @@ name: release-please
 jobs:
   release-please:
     runs-on: ubuntu-latest
+    permissions:
+      contents: write
+      pull-requests: write
     steps:
       - uses: google-github-actions/release-please-action@v4
```
targos, line 14:
```suggestion
      - uses: google-github-actions/release-please-action@v4
        with:
          token: ${{ secrets.GITHUB_TOKEN }}
```

IIRC, I set the token so that the release PR triggers CI workflows.

targos (approved):
Thanks!

.github/workflows/release-please.yml
lukekarrys: Yes I have seen this be an issue before. I thought that `secrets.GITHUB_TOKEN` was the default token so it wouldn't trigger CI workflows, and it would need to pass in a different PAT for that to work.

If this is an issue we can try adding back the token.

targos: You're right. Workflows didn't run for https://github.com/nodejs/gyp-next/pull/218

targos:
https://github.com/nodejs/gyp-next/pull/231

PR #233, opened 2024-03-11
chore: remove component in release tags and set release sha

lukekarrys:
This sets `last-release-sha` to 78756421b0d7bb335992a9c7d26ba3cc8b619708 which is the commit releasing `0.16.2`. This config can be removed anytime in the future after the next release PR is merged.

This also sets `"include-component-in-tag": false` which should make release tags match the old pattern of only including the version prefixed with `v`.

Refs: #232

### DIFF
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -4,7 +4,8 @@
         ".": {
           "release-type": "python",
           "package-name": "gyp-next",
-          "bump-minor-pre-major": true
+          "bump-minor-pre-major": true,
+          "include-component-in-tag": false
         }
     }
 }
--- a/release-please-config.json
+++ b/release-please-config.json
@@ -1,4 +1,5 @@
 {
+    "last-release-sha": "78756421b0d7bb335992a9c7d26ba3cc8b619708",
     "packages": {
         ".": {
           "release-type": "python",


cclauss (approved):

lukekarrys:
Opting for a quick merge here to see if it corrects #232.

lukekarrys:
I think it worked from looking at the workflow logs https://github.com/nodejs/gyp-next/actions/runs/8239001592/job/22531251496

But all the commits since the last release are `chore` so no release PR was created.

But I think 876ccaf5629e1b95e13aaa2b0eb6cbd08fa80593 / #226 should not be a chore? At least from the perspective of `node-gyp` it needs to be released to help fix some linting errors there.

@targos what do you think?

lukekarrys:
If the commit needs to change its message, Release Please allows this by editing the body of the original PR to include a `BEGIN_COMMIT_OVERRIDE` / `END_COMMIT_OVERRIDE` section. This only works if the PR was squash merged though. More info can be found in the docs: https://github.com/googleapis/release-please?tab=readme-ov-file#how-can-i-fix-release-notes

targos:
I wasn't aware that linter issues in this project may affect another one. Added the override section.

targos:
https://github.com/nodejs/gyp-next/pull/234

lukekarrys:
> I wasn't aware that linter issues in this project may affect another one. Added the override section.

Thanks for the release. On this point, maybe it shouldn't affect node-gyp. Python is not my strong suit at all but there are probably changes we can make so that it doesn't.

PR #238, opened 2024-03-30
Fix compile_commands.json for MSVC

segevfiner:
`generator_output` is `None` when trying to use that in Node.js via `vcbuild.bat`

### DIFF
--- a/pylib/gyp/generator/compile_commands_json.py
+++ b/pylib/gyp/generator/compile_commands_json.py
@@ -109,7 +109,7 @@ def GenerateOutput(target_list, target_dicts, data, params):
         AddCommandsForTarget(cwd, target, params, per_config_commands)
 
     try:
-        # generator_output can be `None`, which means the current working directory
+        # generator_output can be `None` on Windows machines
         output_dir = params["options"].generator_output or os.getcwd()
     except (AttributeError, KeyError):
         output_dir = params["generator_flags"].get("output_dir", "out")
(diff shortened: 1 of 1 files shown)


pylib/gyp/generator/compile_commands_json.py
```diff
@@ -109,7 +109,7 @@ def GenerateOutput(target_list, target_dicts, data, params):
         AddCommandsForTarget(cwd, target, params, per_config_commands)
 
     try:
-        output_dir = params["options"].generator_output
+        output_dir = params["options"].generator_output or os.getcwd()
```
anonrig, line 112: Can you add a comment to here to why `or os.getcwd()` is needed?

cclauss: I believe that the comment on 112 already explains why.

pylib/gyp/generator/compile_commands_json.py
```diff
@@ -109,7 +109,8 @@ def GenerateOutput(target_list, target_dicts, data, params):
         AddCommandsForTarget(cwd, target, params, per_config_commands)
 
     try:
-        output_dir = params["options"].generator_output
+        # generator_output can be `None`, which means the current working directory
```
anonrig, line 112:
```suggestion
        # generator_output can be `None` on Windows machines
```

anonrig (approved):

cclauss (approved):

PR #239, opened 2024-04-01
Fix compile_commands.json output location when generator_output is None

segevfiner:
Followup to #238 after discussion at https://github.com/nodejs/node/pull/52279#discussion_r1546658089

cc @anonrig @cclauss

### DIFF
--- a/pylib/gyp/generator/compile_commands_json.py
+++ b/pylib/gyp/generator/compile_commands_json.py
@@ -112,8 +112,8 @@ def GenerateOutput(target_list, target_dicts, data, params):
     try:
         # generator_output can be `None` on Windows machines, or even not
         # defined in other cases
-        output_dir = params["options"].generator_output
-    except (AttributeError, KeyError):
+        output_dir = params.get("options").generator_output
+    except AttributeError:
         pass
     output_dir = output_dir or params["generator_flags"].get("output_dir", "out")
     for configuration_name, commands in per_config_commands.items():
--- a/pylib/gyp/generator/compile_commands_json.py
+++ b/pylib/gyp/generator/compile_commands_json.py
@@ -108,10 +108,13 @@ def GenerateOutput(target_list, target_dicts, data, params):
         cwd = os.path.dirname(build_file)
         AddCommandsForTarget(cwd, target, params, per_config_commands)
 
+    output_dir = None
     try:
-        # generator_output can be `None` on Windows machines
-        output_dir = params["options"].generator_output or os.getcwd()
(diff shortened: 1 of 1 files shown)


pylib/gyp/generator/compile_commands_json.py
```diff
@@ -108,10 +108,14 @@ def GenerateOutput(target_list, target_dicts, data, params):
...
     try:
-        # generator_output can be `None` on Windows machines
-        output_dir = params["options"].generator_output or os.getcwd()
+        # generator_output can be `None` on Windows machines, or even not
+        # defined in other cases
+        output_dir = params["options"].generator_output
     except (AttributeError, KeyError):
+        pass
+    if output_dir is None:
         output_dir = params["generator_flags"].get("output_dir", "out")
```
cclauss, line 119: All other falsely values ("", (), {}, 0, 0.0, etc.) are equally dangerous as `None`
```suggestion
    output_dir = output_dir or params["generator_flags"].get("output_dir", "out")
```

pylib/gyp/generator/compile_commands_json.py
```diff
@@ -108,10 +108,14 @@ def GenerateOutput(target_list, target_dicts, data, params):
...
         AddCommandsForTarget(cwd, target, params, per_config_commands)
 
+    output_dir = None
     try:
-        # generator_output can be `None` on Windows machines
-        output_dir = params["options"].generator_output or os.getcwd()
+        # generator_output can be `None` on Windows machines, or even not
+        # defined in other cases
+        output_dir = params["options"].generator_output
     except (AttributeError, KeyError):
```
cclauss, line 116:
```suggestion
        output_dir = params.get("options").generator_output
    except AttributeError:
```

pylib/gyp/generator/compile_commands_json.py
segevfiner: As you, though I wonder why would someone set it to be an empty dict or some other strange falsey value. 😛

pylib/gyp/generator/compile_commands_json.py
segevfiner: We can also go with `getattr` if you really prefer the optionally get instead of handle exception approach.

StefanStojanovic (approved):
@segevfiner I've checked this code together with https://github.com/nodejs/node/pull/52279 in the node repo, and the build passed. The files were created in `out\{Debug|Release}` as expected.

cclauss (approved):

PR #245, opened 2024-04-27
fix: support empty dictionary keys in input

targos:
Certain old keys in `msvs_settings` (like "LinkIncremental") are moved
to the top level when translated to `msbuild_settings`.
This change allows to set them when `msbuild_settings` are
authored directly.

### DIFF
--- a/pylib/gyp/input.py
+++ b/pylib/gyp/input.py
@@ -2536,6 +2536,8 @@ def ProcessListFiltersInDict(name, the_dict):
     lists = []
     del_lists = []
     for key, value in the_dict.items():
+        if key == "":
+            continue
         operation = key[-1]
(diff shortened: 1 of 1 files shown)


pylib/gyp/input.py
```diff
@@ -2536,6 +2536,8 @@ def ProcessListFiltersInDict(name, the_dict):
     lists = []
     del_lists = []
     for key, value in the_dict.items():
+        if key == "":
```
cclauss, line 2539:
```suggestion
        if not key:
```
All falsey dict keys are just an unwanted as the empty string.
`d = {"": 0, None: 1, False: 2, (): 3, frozenset(): 4, 0: 5, 0.0: 6}`

cclauss (approved):

PR #214, opened 2023-10-20
chore (main): vendor in Python packaging for Py3.12

cclauss:
Fixes #211
Fixes #213

Related to:
* #211
* #213
* nodejs/node-gyp#2862
* nodejs/node-gyp#2869
* nodejs/node-gyp#2888
* nodejs/node-gyp#2912
* nodejs/node-gyp#2915

`distutils` was removed in Python 3.12 and the Python Packaging Authority recommends using
https://github.com/pypa/packaging (https://pypi.org/project/packaging) instead.

Let's _vendor in_ `packaging` so that installing `node-gyp` does not require running `pip install packaging`.

@rzhao271 @DeeDeeG @StefanStojanovic @brjsp

PR #247, opened 2024-04-30
fix: declare packaging in package discovery

legendecas:
`packaging` needs to be declared in setuptools package discovery config.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -118,4 +118,4 @@ max-statements = 286
 
 [tool.setuptools]
 package-dir = {"" = "pylib"}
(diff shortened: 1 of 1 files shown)


legendecas:
yeah, this seems to publish `packaging` globally. I dont think this is the correct way.

legendecas:
@cclauss Would you mind creating a revert on #214? I think `packaging` should be declared as a required dependency, it can be installed on python < 3.12 as well.

cclauss:
The preferred solution is `setuptools; python_version >= "3.12"`

Issue #249, opened 2024-05-05
The return of `ModuleNotFoundError: No module named 'packaging'` on Py3.12


cclauss:
These two commands __fail__ on Python 3.12 with `ModuleNotFoundError: No module named 'packaging'`
% `pipx run --spec git+https://github.com/nodejs/gyp-next.git gyp --help`
% `pipx run --spec git+https://github.com/nodejs/gyp-next.git gyp --version`

This was fixed in the past by:
* #214

Also discussed in:
* #247

PR #250, opened 2024-05-05
chore: pyproject.toml: dependencies = ["packaging>=24.0"]

cclauss:
Fixes: #249 
* #249

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,8 +12,7 @@ description = "A fork of the GYP build system for use in the Node.js projects"
 readme = "README.md"
 license = { file="LICENSE" }
 requires-python = ">=3.8"
-# The Python module "packaging" is vendored in the "pylib/packaging" directory to support Python >= 3.12.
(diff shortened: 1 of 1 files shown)


pyproject.toml
```diff
@@ -12,8 +12,7 @@ description = "A fork of the GYP build system for use in the Node.js projects"
 readme = "README.md"
 license = { file="LICENSE" }
 requires-python = ">=3.8"
-# The Python module "packaging" is vendored in the "pylib/packaging" directory to support Python >= 3.12.
-# dependencies = ["packaging>=23.1"]  # Uncomment this line if the vendored version is removed.
```
legendecas, line 16: Should we remove the vendored `packaging` source code as well?

cclauss: My preference would be to do this in two steps.

legendecas (approved):

legendecas:
I think it would be still necessary to have the vendored packaging source to be used in the nodejs core repo with python 3.12 support. LGTM

Issue #31, opened 2020-04-03
Test integration with Node.js and node-gyp with GH actions


targos:
I think we should make separate workflow files so they can be restarted independently.

For Node.js:
- Clone nodejs/node
- Replace `tools/gyp` with the code from the PR/branch being tested
- Check that Node can be built

For node-gyp:
- clone nodejs/node-gyp
- Replace `gyp` with the code from the PR/branch being tested
- Check that tests pass

PR #255, opened 2024-05-09
ci: add nodejs integration on macos/linux

legendecas:
Add node-integration test to ensure the vendored `packaging` will not be removed by accident.

Refs: https://github.com/nodejs/gyp-next/pull/250#issuecomment-2097070359
Fixes: https://github.com/nodejs/gyp-next/issues/31


.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,67 @@
...
+  pull_request:
+    branches: [ main ]
+  workflow_dispatch:
+
+jobs:
+  integration:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-latest, macos-14, ubuntu-latest]
```
targos, line 14: macos-latest and macos-14 are exactly the same:
https://github.com/nodejs/gyp-next/actions/runs/9015778952/job/24771097889?pr=255
https://github.com/nodejs/gyp-next/actions/runs/9015778952/job/24771097430?pr=255

See "Set up job" -> "Operating System" and "Runner Image"

legendecas: Ah, thanks for pointing this out. I copied this from https://github.com/nodejs/gyp-next/blob/main/.github/workflows/node-gyp.yml#L13. I'll remove this duplication.

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,67 @@
...
+        python: ["3.8", "3.10", "3.12"]
+        exclude:
+          - os: macos-14
+            python: "3.8"
```
cclauss, line 18: This `exclude` is no longer needed.
```suggestion
```

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,67 @@
...
+        run: |
+          rm -rf node/tools/gyp
+          cp -r gyp-next node/tools/gyp
+      - name: Run configure
+        shell: bash
+        run: |
+          cd node
+          ./configure
+
+  integration-windows:
```
cclauss, line 46: These Windows jobs take a long time!

legendecas: I think it could be enough to run the configure step instead of running all the build. In this way it could be much faster.

legendecas: Updated to run `configure` only.

legendecas:
> Why merge all this complexity into a single file?

Most steps are similar in the integration tests, except from platform differences.

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,67 @@
...
+    branches: [ main ]
+  workflow_dispatch:
+
+jobs:
+  integration:
+    strategy:
+      fail-fast: false
+      matrix:
+        os: [macos-latest, macos-14, ubuntu-latest]
+        python: ["3.8", "3.10", "3.12"]
```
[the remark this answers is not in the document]
legendecas: https://github.com/nodejs/node/blob/main/configure#L27 doesn't accept 3.13 yet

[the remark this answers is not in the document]
cclauss: Yes.  Please add it.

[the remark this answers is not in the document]
legendecas: Submitted https://github.com/nodejs/node/pull/53190

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,60 @@
...
+      # macOS and Linux
+      - name: Replace gyp in Node.js
+        if: runner.os != 'Windows'
+        shell: bash
+        run: |
+          rm -rf node/tools/gyp
+          cp -r gyp-next node/tools/gyp
+      - name: Run configure
+        if: runner.os != 'Windows'
+        shell: bash
```
cclauss, line 42: Not needed because there is no Windows.
```suggestion
```

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,60 @@
...
+          repository: nodejs/node
+          path: node
+      - uses: actions/setup-python@v5
+        with:
+          python-version: ${{ matrix.python }}
+          allow-prereleases: true
+
+      # macOS and Linux
+      - name: Replace gyp in Node.js
+        if: runner.os != 'Windows'
```
cclauss, line 35: Bash is the default on Linux and macOS so the `shell: bash` line is only useful on Windows.  This command will work on all three platforms with the `shell: bash` line.
```suggestion
```

.github/workflows/nodejs.yml
```diff
@@ -0,0 +1,60 @@
...
+
+      # Windows
+      - name: Install deps
+        if: runner.os == 'Windows'
+        run: choco install nasm
+      - name: Replace gyp in Node.js
+        if: runner.os == 'Windows'
+        run: |
+          rm -Recurse node/tools/gyp
+          cp -Recurse gyp-next node/tools/gyp
```
cclauss, line 55: This will be done above.
```suggestion
```

cclauss (commented):
Minor optimization

.github/workflows/nodejs.yml
legendecas: Updated, thank you for the suggestion!

cclauss (approved):

PR #256, opened 2024-05-09
chore: dependencies = setuptools for `gyp --version` on Python >= 3.12

cclauss:
Fixes #259

On Python >= 3.12

% `pipx run --no-cache --spec git+https://github.com/nodejs/gyp-next.git gyp --help` was erroring with
> ModuleNotFoundError: No module named 'packaging'

% `pipx run --no-cache --spec git+https://github.com/nodejs/gyp-next.git gyp --version` is still erroring with
> ModuleNotFoundError: No module named 'pkg_resources'

`gyp --help` was fixed in gyp-next v0.18.0
`gyp --version` was not

Since ___both modules___ are included `setuptools` let's use that recommended fix.

---
Proof:
% `pipx install gyp-next`
% `gyp --version ` # Fails!
% `pipx inject gyp-next setuptools ` # This is the same action as this pull request.
% `gyp --version ` # Works as expected.
% `gyp --help ` # Works as expected.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ description = "A fork of the GYP build system for use in the Node.js projects"
 readme = "README.md"
 license = { file="LICENSE" }
 requires-python = ">=3.8"
-dependencies = ["packaging>=24.0"]
+dependencies = ["setuptools>=69.5.1"]
 classifiers = [
     "Development Status :: 3 - Alpha",
     "Environment :: Console",
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -12,7 +12,7 @@ description = "A fork of the GYP build system for use in the Node.js projects"
 readme = "README.md"
 license = { file="LICENSE" }
 requires-python = ">=3.8"
-dependencies = ["setuptools>=69.5.1"]
+dependencies = ["packaging>=24.0", "setuptools>=69.5.1"]
 classifiers = [
     "Development Status :: 3 - Alpha",
     "Environment :: Console",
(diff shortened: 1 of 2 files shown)


ryzokuken (approved):

pyproject.toml
```diff
@@ -12,7 +12,7 @@ description = "A fork of the GYP build system for use in the Node.js projects"
 readme = "README.md"
 license = { file="LICENSE" }
 requires-python = ">=3.8"
-dependencies = ["packaging>=24.0"]
+dependencies = ["setuptools>=69.5.1"]
```
legendecas, line 15: I believe `packaging` should not be removed as installing `setuptools` doesn't come with `packaging`.

```suggestion
dependencies = ["packaging>=24.0", "setuptools>=69.5.1"]
```

legendecas: Downloading setuptools-69.5.1-py3-none-any.whl from https://pypi.org/project/setuptools/#files and unarchiving it, I didn't find that `packaing` is included in `setuptools`.

cclauss: https://github.com/search?q=repo%3Apypa%2Fsetuptools%20packaging&type=code

legendecas: Yes, `packaging` is vendored in setuptools. But as far as I could tell, `packaging` in setuptools is not exposed in `site-packages` with setuptools. So it would need to import like `from setuptools._vendor.packaging.version import Version`.

legendecas: @cclauss I created a github action to verify this change: https://github.com/legendecas/gyp-next/actions/runs/9127187853

legendecas (approved):

PR #259, opened 2024-05-17
chore: add pipx integration test

legendecas:
Refs: https://github.com/nodejs/gyp-next/pull/256

Verify that the built product can work with pipx.


cclauss:
I cherry-picked this into #256
* ##256

PR #264, opened 2024-07-01
chore: fix XCConfigurationList assertion

legendecas:
Fix tests. 

Weirdly, GitHub action didn't run for https://github.com/nodejs/gyp-next/pull/263 and it showed green to be merged. I enabled the following settings for the repo to avoid accidental merge if GitHub action was not triggered.

- Require status checks to pass before merging, and
- Require branches to be up to date before merging (the original PR needs to be updated with the main branch before merging so that we don't have to fixup again).

### DIFF
--- a/.github/workflows/node-gyp.yml
+++ b/.github/workflows/node-gyp.yml
@@ -1,12 +1,11 @@
 name: node-gyp integration
 on:
   push:
-    branches: [ main ]
   pull_request:
-    branches: [ main ]
   workflow_dispatch:
+
 jobs:
-  integration:
+  node-gyp-integration:
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/nodejs.yml
+++ b/.github/workflows/nodejs.yml
@@ -1,13 +1,11 @@
 name: Node.js integration
 on:
   push:
-    branches: [ main ]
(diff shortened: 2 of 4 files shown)


cclauss:
Either line 7 or line 9 (I never remember which) of `.github/workflows/python_tests.yml` should be removed so that GHA's run for PRs from project maintainers.

.github/workflows/node-gyp.yml
```diff
@@ -1,12 +1,11 @@
...
 on:
   push:
-    branches: [ main ]
   pull_request:
-    branches: [ main ]
   workflow_dispatch:
+
 jobs:
-  integration:
+  node-gyp-integration:
```
legendecas, line 8: Renamed with prefix so that this can be cleared identified from merge requirements.

legendecas:
@cclauss would you mind taking a look at this so that the CI on PRs like https://github.com/nodejs/gyp-next/pull/266 can be fixed?

cclauss (approved):
My bet is that the branches changes are going to lead to duplicate builds for maintainers but let's give it a whirl...

legendecas:
I usually push to my own fork repo so I'd bet it won't be triggered for my branches.

PR #270, opened 2024-11-25
ruff rule PLC0206 and RUF021

cclauss:
% [`ruff rule PLC0206`](https://docs.astral.sh/ruff/rules/dict-index-missing-items)
# dict-index-missing-items (PLC0206)

Derived from the **Pylint** linter.

## What it does
Checks for dictionary iterations that extract the dictionary value
via explicit indexing, instead of using `.items()`.

## Why is this bad?
Iterating over a dictionary with `.items()` is semantically clearer
and more efficient than extracting the value with the key.

## Example
```python
ORCHESTRA = {
    "violin": "strings",
    "oboe": "woodwind",
    "tuba": "brass",
    "gong": "percussion",
}

for instrument in ORCHESTRA:
    print(f"{instrument}: {ORCHESTRA[instrument]}")
```

Use instead:
```python
ORCHESTRA = {
    "violin": "strings",
    "oboe": "woodwind",
    "tuba": "brass",
    "gong": "percussion",
}

for instrument, section in ORCHESTRA.items():
    print(f"{instrument}: {section}")
```
---
% [`ruff rule RUF021`](https://docs.astral.sh/ruff/rules/parenthesize-chained-operators)
# parenthesize-chained-operators (RUF021)

Derived from the **Ruff-specific rules** linter.

Fix is always available.

## What it does
Checks for chained operators where adding parentheses could improve the
clarity of the code.

## Why is this bad?
`and` always binds more tightly than `or` when chaining the two together,
but this can be hard to remember (and sometimes surprising).
Adding parentheses in these situations can greatly improve code readability,
with no change to semantics or performance.

For example:
```python
a, b, c = 1, 0, 2
x = a or b and c

d, e, f = 0, 1, 2
y = d and e or f
```

Use instead:
```python
a, b, c = 1, 0, 2
x = a or (b and c)

d, e, f = 0, 1, 2
y = (d and e) or f
```

### DIFF
--- a/pylib/gyp/MSVSSettings.py
+++ b/pylib/gyp/MSVSSettings.py
@@ -171,7 +171,7 @@ class _Integer(_Type):
         int(value, self._msbuild_base)
 
     def ConvertToMSBuild(self, value):
-        msbuild_format = (self._msbuild_base == 10) and "%d" or "0x%04x"
+        msbuild_format = ((self._msbuild_base == 10) and "%d") or "0x%04x"
         return msbuild_format % int(value)
 
 
--- a/pylib/gyp/MSVSVersion.py
+++ b/pylib/gyp/MSVSVersion.py
@@ -69,7 +69,7 @@ class VisualStudioVersion:
 
     def ProjectExtension(self):
         """Returns the file extension for the project."""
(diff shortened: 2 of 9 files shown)


cclauss:
@nodejs/python

PR #271, opened 2024-12-02
feat: provide escaped version of `PRODUCT_DIR_ABS`

addaleax:
Since Node.js uses this in a C string, we should provide a version that escapes it as a C string.

Needed by: https://github.com/nodejs/node/pull/56111

### DIFF
--- a/pylib/gyp/__init__.py
+++ b/pylib/gyp/__init__.py
@@ -4,7 +4,7 @@
 # Use of this source code is governed by a BSD-style license that can be
 # found in the LICENSE file.
 
-
+from __future__ import annotations
 import copy
 import gyp.input
 import argparse
--- a/pylib/gyp/__init__.py
+++ b/pylib/gyp/__init__.py
@@ -24,6 +24,17 @@ DEBUG_GENERAL = "general"
 DEBUG_VARIABLES = "variables"
 DEBUG_INCLUDES = "includes"
 
+def EscapeForCString(string):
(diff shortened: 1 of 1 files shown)


pylib/gyp/__init__.py
```diff
@@ -24,6 +24,17 @@
 DEBUG_VARIABLES = "variables"
 DEBUG_INCLUDES = "includes"
 
+def EscapeForCString(string):
+    if isinstance(string, str):
+        string = string.encode(encoding='utf8')
+
+    result = ''
+    for char in string:
+        if not (32 <= char < 127) or char in (ord('\\'), ord('"')):
```
cclauss, line 33: Type hints and set lookup is superfast in Python.
```suggestion
def EscapeForCString(string: bytes | str) -> str:
    if isinstance(string, str):
        string = string.encode(encoding='utf8')

    backslash_or_double_quote = {ord('\\'), ord('"')}
    result = []
    for char in string:
        if char in backslash_or_double_quote or not 32 <= char < 127:
```

cclauss (approved):

pylib/gyp/__init__.py
addaleax: @cclauss thank you, updated! 🙂

addaleax:
@cclauss sure, done! 🙇‍♀️

anonrig (approved):

PR #274, opened 2024-12-09
fix: fixup for break in EscapeForCString

addaleax:
The review suggestion in #271 accidentally changed the type of `result` and broke this part.


cclauss (approved):

addaleax:
@cclauss I don't think I have permissions to release here, could you try to do that?

PR #278, opened 2024-12-28
Remove explicit installation of setuptools

### LINKED ISSUE #277: At some point, we should stop telling contributors to update `setuptools`
[`docs/Hacking.md` says:](https://github.com/nodejs/gyp-next/blob/a61df708e3c2249f54b32e355598b11e35e7612e/docs/Hacking.md#testing-your-change)

> GYP has a suite of tests which you can run with the provided test driver
> to make sure your changes aren't breaking anything important.
>
> You run the test driver with e.g.
>
> ``` sh
> $ python -m pip install --upgrade pip setuptools
> […]
> ```

[As of Python 3.12, Python no longer comes with the `setuptools` package by default.](https://docs.python.org/3/whatsnew/3.12.html#ensurepip) I tried following the instructions in `docs/Hacking.md` in a fresh Python 3.13 virtual environment, but I skipped installing/upgrading `setuptools` (I only upgraded `pip`). Everything worked fine.

At some point, `docs/Hacking.md` so that it doesn’t tell users to upgrade `setuptools`. We could do it after Python 3.10 has reached end of life, or we could do it sooner with a note that says “hey, if you’re using Python 3.12+, don’t upgrade `setuptools`”.

cclauss:
Fixes #277
* #277

@Jayman2000

### DIFF
--- a/pylib/gyp/__init__.py
+++ b/pylib/gyp/__init__.py
@@ -696,7 +696,7 @@ def main(args):
         return 1
 
 
-# NOTE: setuptools generated console_scripts calls function with no arguments
+# NOTE: console_scripts calls this function with no arguments
 def script_main():
     return main(sys.argv[1:])
 
--- a/.github/workflows/node-gyp.yml
+++ b/.github/workflows/node-gyp.yml
@@ -33,7 +33,7 @@ jobs:
       - name: Install Python dependencies
         run: |
           cd gyp-next
-          python -m pip install --upgrade pip setuptools
+          python -m pip install --upgrade pip
           pip install --editable .
           pip uninstall -y gyp-next
       - name: Install Node.js dependencies
--- a/.github/workflows/python_tests.yml
+++ b/.github/workflows/python_tests.yml
@@ -26,7 +26,7 @@ jobs:
       - uses: seanmiddleditch/gha-setup-ninja@v5
       - name: Install dependencies
         run: |
-          python -m pip install --upgrade pip setuptools
+          python -m pip install --upgrade pip
           pip install --editable ".[dev]"
       - run: ./gyp -V && ./gyp --version && gyp -V && gyp --version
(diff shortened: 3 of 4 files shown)


pylib/gyp/__init__.py
```diff
@@ -696,7 +696,7 @@ def main(args):
         return 1
 
 
-# NOTE: setuptools generated console_scripts calls function with no arguments
+# NOTE: console_scripts calls this function with no arguments
```
legendecas, line 699: I think the original comment is still correct and concise as this function is referenced in `pyproject.toml#project.scripts`?

Jayman2000 (approved):
I created [a script that tests this PR in Python 3.8 through Python 3.13](https://gist.github.com/Jayman2000/b21ab8962da0a4eb87171de4d9434f8b). Everything seems to work fine in all of those versions of Python.

pylib/gyp/__init__.py
cclauss: The Python Packaging Authortiy has respecified `console_scripts` to be supported beyond `setuptools` as discussed at
https://packaging.python.org/en/latest/specifications/entry-points

cclauss:
@Jayman2000 When you think a pull request is useful and is ready to be merged, ___please consider giving it a positive review___.

Every check mark ✔️ at the top right of this page gives project maintainers confidence that the proposed changes have been read through and deemed both useful and safe to merge into the codebase. 

Anyone can review a pull request on GitHub.  To do so here:
1. Scroll to the top of this page.
2. Click the `Files changed` tab and read through each file carefully looking for potential issues.
3. Click the `Review changes` button.
4. Click `Approve` (or one of the other options) and add comments only if they have not already been stated in the PR.
5. Click `Submit review` so that your ✔️ will be added to the list.

legendecas (approved):

PR #291, opened 2025-03-24
fix: python lint  more ruff rules

cclauss:
% `ruff check --select=ALL --ignore=COM,D,Q,RET,SIM910 --statistics | grep "\[\*\]" | sort -k2`
```
   2	FURB105	[*] print-empty-string
  36	I001   	[*] unsorted-imports
  14	ISC001 	[*] single-line-implicit-string-concatenation
   2	PIE790 	[*] unnecessary-placeholder
   1	PIE808 	[*] unnecessary-range-start
  19	PLR0402	[*] manual-from-import
   8	PLR5501	[*] collapsible-else-if
   6	SIM114 	[*] if-with-same-arms
   6	SIM300 	[*] yoda-conditions
[*] 94 fixable with the `--fix` option
```
% `ruff rule I001`
# unsorted-imports (I001)

Derived from the **isort** linter.

Fix is sometimes available.

## What it does
De-duplicates, groups, and sorts imports based on the provided `isort` settings.

## Why is this bad?
Consistency is good. Use a common convention for imports to make your code
more readable and idiomatic.

## Example
```python
import pandas
import numpy as np
```

Use instead:
```python
import numpy as np
import pandas
```


pylib/gyp/msvs_emulation.py
```diff
@@ -7,15 +7,15 @@
 build systems, primarily ninja.
 """
 
-import collections
 import os
 import re
 import subprocess
 import sys
+from collecttions import namedtuple
```
cclauss, line 14:
```suggestion
from collections import namedtuple
```

legendecas (approved):

PR #293, opened 2025-04-12
fix: Ensure Consistent Order of build_files in WriteAutoRegenerationRule

zhan9san:
Related to https://github.com/nodejs/node-gyp/pull/3062. Thank @ben-zalekta-lmnd

This PR addresses an issue with the WriteAutoRegenerationRule function in node-gyp, where the build_files set was passed without sorting, leading to an inconsistent order of dependencies in the generated Makefile.

Changes:

Wrapped the build_files set in the sorted() function before it is processed in WriteAutoRegenerationRule. This ensures that the order of files is consistent across all runs, preventing issues in CI environments that rely on file hash comparisons.
Impact:
This change stabilizes the output of the Makefile generation process, resolving the issue of inconsistent file order and reducing false positives in CI checks.

### DIFF
--- a/pylib/gyp/generator/make.py
+++ b/pylib/gyp/generator/make.py
@@ -2381,7 +2381,7 @@ def WriteAutoRegenerationRule(params, root_makefile, makefile_name, build_files)
         % {
             "makefile_name": makefile_name,
             "deps": replace_sep(
-                " ".join(SourceifyAndQuoteSpaces(bf) for bf in build_files)
+                " ".join(sorted(SourceifyAndQuoteSpaces(bf) for bf in build_files))
             ),
             "cmd": replace_sep(gyp.common.EncodePOSIXShellList(
                 [gyp_binary, "-fmake"] + gyp.RegenerateFlags(options) + build_files_args


zhan9san:
@lukekarrys 

Could you help review this PR?

pylib/gyp/generator/make.py
```diff
@@ -2381,7 +2381,7 @@ def WriteAutoRegenerationRule(params, root_makefile, makefile_name, build_files)
         % {
             "makefile_name": makefile_name,
             "deps": replace_sep(
-                " ".join(SourceifyAndQuoteSpaces(bf) for bf in build_files)
+                " ".join(SourceifyAndQuoteSpaces(bf) for bf in sorted(build_files))
```
cclauss, line 2384:
```suggestion
                " ".join(sorted( SourceifyAndQuoteSpaces(bf) for bf in build_files))
```
Would `SourceifyAndQuoteSpaces(build_file)` change the sort order of the files?

zhan9san: @cclauss 

Thanks for your review.

I don't think `SourceifyAndQuoteSpaces` would change the sort order. Instead, it just format the path as below.
https://github.com/nodejs/gyp-next/blob/e97fd2509a6fb6452eab5e9b17f4cae3fc49f25e/pylib/gyp/generator/make.py#L749-L764

zhan9san:
@cclauss 

Just a gentle reminder to see if you’ve had a chance to review my PR — happy to make any changes if needed!

PR #311, opened 2025-08-31
fix: handle `None` case in xcode_emulation regexes

aduh95:
In the case where `softwareupdate --history` does not list any CLT update, the program crashes with an unhelpful error instead of trying the next fallback.

### DIFF
--- a/pylib/gyp/xcode_emulation.py
+++ b/pylib/gyp/xcode_emulation.py
@@ -1534,18 +1534,22 @@ def CLTVersion():
     FROM_XCODE_PKG_ID = "com.apple.pkg.DeveloperToolsCLI"
     MAVERICKS_PKG_ID = "com.apple.pkg.CLTools_Executables"
 
-    regex = re.compile("version: (?P<version>.+)")
+    regex = re.compile(r"version: (?P<version>.+)")
     for key in [MAVERICKS_PKG_ID, STANDALONE_PKG_ID, FROM_XCODE_PKG_ID]:
         try:
             output = GetStdout(["/usr/sbin/pkgutil", "--pkg-info", key])
-            return re.search(regex, output).groupdict()["version"]
+            m = re.search(regex, output)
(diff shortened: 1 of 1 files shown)


pylib/gyp/xcode_emulation.py
```diff
@@ -1534,18 +1534,22 @@ def CLTVersion():
...
     MAVERICKS_PKG_ID = "com.apple.pkg.CLTools_Executables"
 
-    regex = re.compile("version: (?P<version>.+)")
+    regex = re.compile(r"version: (?P<version>.+)")
     for key in [MAVERICKS_PKG_ID, STANDALONE_PKG_ID, FROM_XCODE_PKG_ID]:
         try:
             output = GetStdout(["/usr/sbin/pkgutil", "--pkg-info", key])
-            return re.search(regex, output).groupdict()["version"]
+            m = re.search(regex, output)
+            if m:
```
cclauss, line 1542:
```suggestion
            if m := re.search(regex, output):
```

pylib/gyp/xcode_emulation.py
```diff
@@ -1534,18 +1534,22 @@ def CLTVersion():
...
+                return m.groupdict()["version"]
         except (GypError, OSError):
             continue
 
     regex = re.compile(r"Command Line Tools for Xcode\s+(?P<version>\S+)")
     try:
         output = GetStdout(["/usr/sbin/softwareupdate", "--history"])
-        return re.search(regex, output).groupdict()["version"]
+        m = re.search(regex, output)
+        if m:
```
cclauss, line 1551:
```suggestion
        if m := re.search(regex, output):
```

cclauss (changes requested):
Nice!

Can you please add a sample of the current error stack trace?

Would adding a pytest be possible so we avoid regressions?

legendecas (approved):

PR #339, opened 2026-04-15
Drop support for Python 3.8

cclauss:
Both Python 3.8 and 3.9 are [end-of-life](https://devguide.python.org/versions), but we might want to continue to support Py3.9 until the [31 August 2026 EOL](https://wiki.debian.org/DebianReleases#Versions_of_Debian_stable) of Debian 11 Bullseye Standard Support.

### DIFF
--- a/.github/workflows/python_tests.yml
+++ b/.github/workflows/python_tests.yml
@@ -8,10 +8,7 @@ on:
 
 jobs:
   Python_lint:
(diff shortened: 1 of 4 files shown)


aduh95 (approved):
I'd be fine dropping 3.9 as well (and potentially wait until September to release it), but I'm also good with the current approach

PR #355, opened 2026-08-24
fix: write pch cflags as a list instead of a map object

PickBas:
## Problem

`GetFlagsModifications` returned `map` objects for `cflags_c`/`cflags_cc`.

The `/Yc` flag never reached `cl.exe`, so precompiled headers were broken for every ninja build on Windows.

## Fix

Replace both `map()` calls with list comprehensions.

## Test

`test_ninja_precompiled_header` runs the ninja generator over a new PCH fixture and asserts the flags are written out instead of a `map` object.

Refs: https://github.com/nodejs/node/issues/57633

### DIFF
--- a/pylib/gyp/msvs_emulation.py
+++ b/pylib/gyp/msvs_emulation.py
@@ -1057,13 +1057,18 @@ class PrecompiledHeader:
             pch_output = ["/Yc" + self._PchHeader()]
             if command == "cxx":
                 return (
-                    [("cflags_cc", map(expand_special, cflags_cc + pch_output))],
+                    [
+                        (
+                            "cflags_cc",
+                            [expand_special(x) for x in cflags_cc + pch_output],
+                        )
+                    ],
                     self.output_obj,
                     [],
                 )
             elif command == "cc":
                 return (
-                    [("cflags_c", map(expand_special, cflags_c + pch_output))],
+                    [("cflags_c", [expand_special(x) for x in cflags_c + pch_output])],
                     self.output_obj,
                     [],
                 )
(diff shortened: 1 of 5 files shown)


pylib/gyp/msvs_emulation.py
```diff
@@ -1057,13 +1057,13 @@ def GetFlagsModifications(
             pch_output = ["/Yc" + self._PchHeader()]
             if command == "cxx":
                 return (
-                    [("cflags_cc", map(expand_special, cflags_cc + pch_output))],
+                    [("cflags_cc", list(map(expand_special, cflags_cc + pch_output)))],
```
cclauss:
```suggestion
                    [("cflags_cc", [expand_special(x) for x in cflags_cc + pch_output])],
```
Use a list comprehension as recommended in "What's new in Python 3.0":
https://docs.python.org/3/whatsnew/3.0.html#views-and-iterators-instead-of-lists

PickBas: @cclauss Changed in both places, thank you

test/integration_test.py
```diff
@@ -12,6 +12,7 @@
 
 fixture_dir = os.path.join(os.path.dirname(__file__), "fixtures")
 gyp_file = os.path.join(os.path.dirname(__file__), "fixtures/integration.gyp")
+pch_gyp_file = os.path.join(os.path.dirname(__file__), "fixtures/pch.gyp")
```
cclauss:
```suggestion
fixture_dir = os.path.join(os.path.dirname(__file__), "fixtures")
gyp_file = os.path.join(fixture_dir, "integration.gyp")
pch_gyp_file = os.path.join(fixture_dir, "pch.gyp")
```

PickBas: @cclauss Fixed, thank you

cclauss (approved):
LGTM

StefanStojanovic (approved):

